PR #3, opened 2017-04-06
Update testing framework.

schlessera:


schlessera:
Package seems to include custom code right now, aborting PR.

PR #7, opened 2017-04-17
Generate 24 character password for new users

tillkruss:
This matches the default password length for new users.

See also: https://github.com/wp-cli/wp-cli/pull/4002


tillkruss:
Thanks!

PR #8, opened 2017-04-18
Add `wp option` from wp-cli/wp-cli

danielbachhuber:
From
https://github.com/wp-cli/wp-cli/commit/3bbe5ff075ff106484789186a35d3cca03b02758


danielbachhuber:
After this lands, need to tag a new version, and then update https://github.com/wp-cli/wp-cli/pull/4007

PR #21, opened 2017-06-29
Support emails in user_login field

soulseekah:
Logins can be set to e-mails, and are often set this way. This is fine,
until the user changes the e-mail (logins can't be changed). What
happens then? `wp user get` doesn't find the user by login, because if
the login is an e-mail it shorts to checking only the user_email field.

Rectify this. Tests included.


soulseekah:
Thanks :)

PR #24, opened 2017-07-05
Add pluck and patch subcommands

aaemnnosttv:
Hey guys,

I thought I would post this here for some feedback before I go on and do much more.

I've started with the meta commands as it seemed like the easiest way to affect the most commands.

I added some things like PHPUnit which helped a lot with testing the class I added; that OK?

I expect some things will need to change but let me know if we're on the right track here, or if you'd like me to be doing something different while things are more flexible.

Resolves wp-cli/ideas#42


features/post-meta.feature
```diff
@@ -139,3 +139,95 @@ Feature: Manage post custom fields
...
+          }
+        },
+        "foo.com": {
+          "visitors": 999
+        }
+      }
+      """
+    And I run `wp post meta set 1 meta-key --format=json < input.json`
+
+    When I run `wp post meta pluck 1 meta-key foo.bar.baz`
```
danielbachhuber: What if this were space limited? e.g.

```
wp post meta pluck 1 meta-key foo bar baz
```

I think space delimited is closer to existing WP-CLI patterns, unless there's a specific reason you're using periods.

features/post-meta.feature
```diff
@@ -139,3 +139,95 @@ Feature: Manage post custom fields
...
+    And I run `wp post meta set 1 meta-key '{ "key": "value" }' --format=json`
+
+    When I run `wp post meta pluck 1 meta-key key`
+    Then STDOUT should be:
+      """
+      value
+      """
+
+    When I try `wp post meta pluck 1 meta-key foo`
+    Then STDOUT should be empty
```
danielbachhuber: We should set an explicit check for return code here too:

```
And the return code should be 1
```

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,127 @@ public function update( $args, $assoc_args ) {
...
+	 * : The name of the inner key to get.
+	 *
+	 * [--delimiter=<character>]
+	 * : The character to delimit hierarchy within the pluck-key.
+	 * ---
+	 * default: .
+	 * ---
+	 *
+	 * [--format=<format>]
+	 * : Accepted values: table, json. Default: table
```
danielbachhuber: Are these formats correct?

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,127 @@ public function update( $args, $assoc_args ) {
...
+	 * : The new value. If omitted, the value is read from STDIN.
+	 *
+	 * [--delimiter=<character>]
+	 * : The character to delimit hierarchy within the patch-key.
+	 * ---
+	 * default: .
+	 * ---
+	 *
+	 * [--mode=<mode>]
+	 * : Configures the behavior of the value update.
```
danielbachhuber: If a user supplies a new value but uses `--mode=unset`, what's the behavior?

Would it make more sense to have a separate method for deleting?

bin/test.sh
```diff
@@ -2,6 +2,8 @@
 
 set -ex
 
+phpunit
```
danielbachhuber: We should make this a part of the default package scaffold too: https://github.com/wp-cli/scaffold-package-command/issues/121

features/post-meta.feature
aaemnnosttv: I thought about that. My initial thought was to use a single string for it and make the delimiter configurable which is how it is now. A space could still be used as it is but it would require wrapping in quotes. The command is a little lengthy with arguments as it is, I think the dot-notation it's a little easier to read.

src/WP_CLI/CommandWithMeta.php
aaemnnosttv: This is mostly copied from the existing `get` subcommand. The table output is a misnomer for sure as it does not output as a table.

It runs through `\WP_CLI::print_value` so it should probably be updated with the options it supports.

src/WP_CLI/CommandWithMeta.php
aaemnnosttv: No test for this yet but it would unset the key which is different than say `wp post meta patch 1 meta-key foo.bar ''` and just setting it to an empty value.

I like the idea of using a separate command but can't think of a good name to use for that. I think that it can work under the umbrella of `patch` as it is still a nested value modification unless we can come up with a verb that makes more sense? It's "patching" the value with the key removed.

I was also thinking that there may be a `merge` mode a merge of data rather than just replacing, but I pulled back on it as it probably wouldn't be very practical.

features/post-meta.feature
schlessera: I think the preferred approach should be to accept an arbitrary number of positional arguments with each own representing one level.

The benefit is that you can have both periods and spaces be part of the keys:
```
# foo -> bar -> baz
wp post meta pluck 1 meta-key foo bar baz

# foo.bar -> baz
wp post meta pluck 1 meta-key foo.bar baz

# foo bar -> baz
wp post meta pluck 1 meta-key "foo bar" baz
```

Any mechanism which needs to to provide all levels in one positional argument will force you to have one arbitrary character be impossible to use as a key identifier.

aaemnnosttv: That sounds like a good idea. For `patch` then, should we use an associative argument for the new value or continue to use the last positional argument as the new value?

danielbachhuber: > For `patch` then, should we use an associative argument for the new value or continue to use the last positional argument as the new value?

Last positional argument seems reasonable to me.

aaemnnosttv: Will the Synopsis parser allow a `<parts...> [<value>]` ?

aaemnnosttv:
Commands have been updated to use positional args as the target key path, and the concept of a delimiter has been removed. I also updated the assertions to use JSON for consistency with the input.

We still need to come to a decision as to how `unset` should be used either as a patch flag or separate command. I found a similar case which we will need to make a similar decision for in the case of wanting to set a value for a non-existent key.

By default, the command errors out if the key does not exist. If we want to create the key with the value (ie. insert it) then we need a way to set that in the command.

This is where I am looking for feedback from you guys.  I'm not a big fan of using/keeping the `mode` flag, but I also don't really like the idea of adding 2 more named subcommands which are slightly different than `patch`, particularly because I can't think of any good names for the operations that wouldn't be confused for other crud commands. `unset` could be confused with `delete`, and something like `insert` could be confused with `add`. Thoughts?

danielbachhuber:
> This is where I am looking for feedback from you guys. I'm not a big fan of using/keeping the `mode` flag, but I also don't really like the idea of adding 2 more named subcommands which are slightly different than `patch`, particularly because I can't think of any good names for the operations that wouldn't be confused for other crud commands.

This is my sentiment too. I don't have a strong opinion one way or another at this point (`--mode=<mode>` argument vs. separate subcommands).

Another idea would be: `wp option patch (add|remove|update|delete)`.

danielbachhuber:
How does this work with numerically indexed values?

aaemnnosttv:
> Another idea would be: wp option patch (add|remove|update|delete).

I like this the most out of all the options so far 👍 

> How does this work with numerically indexed values?

Good question! The argument will be interpreted as a string I believe and will check for a key with a loose comparison. The traverser can obviously handle any valid type as keys to navigate, the limitation is more on taking the input from the command args where the type will have to be inferred. I don't think there will be a perfect solution here, but I think the only situation where it would be problematic is if a key existed for both the string and numeric value.

E.g.
```php
array(
 1 => 'int one',
 '1' => 'string one',
)
```

This seems like an edge case that isn't worth handling, but it is something we should look into and make sure that it works reasonably well with numeric keys.

I'll add some tests for this and keep this in mind when building out `add/insert`.

danielbachhuber:
> I'll add some tests for this and keep this in mind

Great, thanks. I don't think we need a solution for the ambiguous scenario either, but it would be good to have a test with explicit assertions of behavior.

aaemnnosttv:
Update! Pluck and patch are functionally complete for meta commands 🎉  

Patch now requires the action to be set as the first argument `(insert|update|delete)` which works quite nicely!

More tests have been added for everything, including the behavior with indexed keys.

I thought indexed keys would be more of a pain than they were, but it turns out we only needed to cast them to integers (when they look like one) and everything just works. I tried to create a scenario for the edge case where I was expecting this to fail using a string integer for the key, but found this quite hard to do. Both JSON and PHP convert string integers to integers when serializing.

```php
serialize( ["0" => "foo"] )
// string(20) "a:1:{i:0;s:3:"foo";}"
```

Who knew!? 😁

The question now is whether this looks good enough to start replicating to the rest of the entities?
Also, I was thinking about namespaces. Should `\WP_CLI\RecursiveDataStructureTraverser` be moved under `\WP_CLI\Entity\RecursiveDataStructureTraverser` ? I realize that probably can't be done with the other classes in the package since they existed before the split, but it seems like it might be a good idea going forward.

danielbachhuber:
> Update! Pluck and patch are functionally complete for meta commands

Sweet :) Looking good on my end. Given the amount of code landing, I'd like for @schlessera to give it a review too.

> Should `\WP_CLI\RecursiveDataStructureTraverser` be moved under `\WP_CLI\Entity\RecursiveDataStructureTraverser`?

Yes, I think this would be a good idea.

> I realize that probably can't be done with the other classes in the package since they existed before the split, but it seems like it might be a good idea going forward.

Correct, and sounds good.

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,140 @@ public function update( $args, $assoc_args ) {
...
+	 * : The name(s) of the keys within the value to locate the value to pluck.
+	 *
+	 * [--format=<format>]
+	 * : The output format of the value.
+	 * ---
+	 * default: plaintext
+	 * options:
+	 *   - plaintext
+	 *   - json
+	 *   - yaml
```
danielbachhuber: Could we get a couple useful `## EXAMPLES` here?

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,140 @@ public function update( $args, $assoc_args ) {
...
+	 * : The new value. If omitted, the value is read from STDIN.
+	 *
+	 * [--format=<format>]
+	 * : The serialization format for the value.
+	 * ---
+	 * default: plaintext
+	 * options:
+	 *   - plaintext
+	 *   - json
+	 * ---
```
danielbachhuber: Could we get a couple useful `## EXAMPLES` here?

features/post-meta.feature
```diff
@@ -139,3 +139,273 @@ Feature: Manage post custom fields
...
+    Then STDOUT should be:
+      """
+      Success: Updated custom field 'meta-key'.
+      """
+
+    When I run `wp post meta get 1 meta-key --format=json`
+    Then STDOUT should be JSON containing:
+      """
+      [ "new", "bar" ]
+      """
```
schlessera: I think there's three use/test cases missing:

1. What happens when you `patch insert` an already existing value? (Is there a warning? Does it fail?)
2. What happens when you `patch update` a non-existing key? (This should produce an error.)
3. What happens when you `patch update` an existing key to the same value? (You seem to have an adapted success message in the code below.)

phpunit.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/5.7/phpunit.xsd"
```
schlessera: You have PHPUnit ^4.8 in the Composer requirements, not 5.7.

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,140 @@ public function update( $args, $assoc_args ) {
...
+		}, array_slice( $args, 2 ) );
+
+		$value = get_metadata( $this->meta_type, $object_id, $meta_key, true );
+
+		$traverser = new RecursiveDataStructureTraverser( $value );
+
+		try {
+			$value = $traverser->get( $key_path );
+		} catch ( \Exception $e ) {
+			die( 1 );
```
schlessera: Instead of silently dying here, you should use something like the following:

```
WP_CLI::error( "Failed to recursively traverse meta value for key '{$meta_key}'. Reason: {$e->getMessage()}" );
```

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,140 @@ public function update( $args, $assoc_args ) {
...
+
+		/* Need to make a copy of $current_meta_value here as it is modified by reference */
+		$current_meta_value = $old_meta_value = sanitize_meta( $meta_key, get_metadata( $this->meta_type, $object_id, $meta_key, true ), $this->meta_type );
+
+		$traverser = new RecursiveDataStructureTraverser( $current_meta_value );
+
+		try {
+			$traverser->$action( $key_path, $patch_value );
+		} catch ( \Exception $e ) {
+			WP_CLI::error( $e->getMessage() );
```
schlessera: Same as my suggestion above, I would include some context as well as the `$meta_key` that was meant to be acted upon.

This helps debug issues when a script doing bulk changes fails at random points.

src/WP_CLI/CommandWithMeta.php
```diff
@@ -254,6 +254,140 @@ public function update( $args, $assoc_args ) {
...
+		}
+
+		$patched_meta_value = sanitize_meta( $meta_key, $traverser->value(), $this->meta_type );
+
+		if ( $patched_meta_value === $old_meta_value ) {
+			WP_CLI::success( "Value passed for custom field '$meta_key' is unchanged." );
+		} else {
+			$slashed = wp_slash( $patched_meta_value );
```
schlessera: Why the slashing here?

Here's what the Codex says:
```
Metadata value (this should be raw as opposed to sanitized for database queries).
```

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+
+	/**
+	 * @var null|static The parent instance of the traverser.
+	 */
+	protected $parent;
+
+	/**
+	 * RecursiveDataStructureTraverser constructor.
+	 *
+	 * @param        $data
```
schlessera: Typehint missing. Is this `mixed`, or can you provide a more precise typehint?

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	 * @param static $parent
+	 */
+	public function __construct( &$data, $parent = null ) {
+		$this->data =& $data;
+		$this->parent = $parent;
+	}
+
+	/**
+	 * @param $locator
```
schlessera: Typehint missing. `array|string|int` ?

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	/**
+	 * Get the current data.
+	 *
+	 * @return mixed
+	 */
+	public function value() {
+		return $this->data;
+	}
+
+	public function update( $locator, $value ) {
```
schlessera: Docblock for public method needed.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+		$this->traverse_to( (array) $locator )->set_value( $value );
+	}
+
+	public function set_value( $value ) {
+		$this->data = $value;
+	}
+
+	public function delete( $locator ) {
```
schlessera: Docblock for public method needed.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+		$this->traverse_to( (array) $locator )->unset_on_parent();
+	}
+
+	public function insert( $locator, $value ) {
```
schlessera: Docblock for public method needed.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+		try {
+			$this->update( $locator, $value );
```
schlessera: So the `patch insert` falls back to a `patch update` in case the key exists. Should this create a warning? Should it only fall back to `patch update` with something like a `--force` flag?

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	 */
+	public function value() {
+		return $this->data;
+	}
+
+	public function update( $locator, $value ) {
+		$this->traverse_to( (array) $locator )->set_value( $value );
+	}
+
+	public function set_value( $value ) {
```
schlessera: Should this be part of the public interface? If not, it should be `protected`.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+		} catch ( NonExistentKeyException $e ) {
+			$e->get_traverser()->create_key();
+			$this->insert( $locator, $value );
+		}
+	}
+
+	/**
+	 * Delete the key on the parent's data that references this data.
+	 */
+	public function unset_on_parent() {
```
schlessera: Should this be part of the public interface? If not, it should be `protected`.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+		$this->parent->delete_by_key( $this->key );
+	}
+
+	/**
+	 * Delete the given key from the data.
+	 *
+	 * @param $key
+	 */
+	public function delete_by_key( $key ) {
```
schlessera: Should this be part of the public interface? If not, it should be `protected`.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	/**
+	 * Get an instance of the traverser for the given hierarchical key.
+	 *
+	 * @param $locator
+	 *
+	 * @throws \Exception
+	 *
+	 * @return static
+	 */
+	public function traverse_to( $locator ) {
```
schlessera: Should this be part of the public interface? If not, it should be `protected`.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+					$traverser->set_key( $key );
+					return $traverser->traverse_to( $locator );
+				}
+			}
+		}
+
+		return $this;
+	}
+
+	public function set_key( $key ) {
```
schlessera: Should this be part of the public interface? If not, it should be `protected`.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	/**
+	 * Check if the given key exists on the current data.
+	 *
+	 * @param string $key
+	 *
+	 * @return bool
+	 */
+	public function exists( $key ) {
```
schlessera: Should this be part of the public interface? If not, it should be `protected`.

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+			unset( $this->data->$key );
+		}
+	}
+
+	/**
+	 * Get an instance of the traverser for the given hierarchical key.
+	 *
+	 * @param $locator
+	 *
+	 * @throws \Exception
```
schlessera: Should be: `@throws NonExistentKeyException`

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	 */
+	public function __construct( &$data, $parent = null ) {
+		$this->data =& $data;
+		$this->parent = $parent;
+	}
+
+	/**
+	 * @param $locator
+	 *
+	 * @throws \Exception
```
schlessera: Should be: `@throws NonExistentKeyException`

src/WP_CLI/RecursiveDataStructureTraverser.php
```diff
@@ -0,0 +1,159 @@
...
+	 * @internal param string $key
+	 */
+	protected function create_key() {
+		if ( is_array( $this->data ) ) {
+			$this->data[ $this->key ] = null;
+		} elseif ( is_object( $this->data ) ) {
+			$this->data->{$this->key} = null;
+		} else {
+			throw new \Exception( "Cannot create key '$this->key', invalid type." );
```
schlessera: This should be a more specific exception. Either a custom one, like `NonExistentKeyException`is (maybe `InvalidDataTypeException`?) or at least one of the [exceptions provided by the SPL](http://php.net/manual/de/spl.exceptions.php).

In this case, I think an [`UnexpectedValueException`](http://php.net/manual/de/class.unexpectedvalueexception.php) is a good fit.

src/WP_CLI/NonExistentKeyException.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace WP_CLI;
+
+class NonExistentKeyException extends \Exception {
```
schlessera: This should extend one of the [exceptions provided by the SPL](http://php.net/manual/de/spl.exceptions.php).

I think the [`OutOfBoundsException`](http://php.net/manual/de/class.outofboundsexception.php) should be a good fit.

tests/RecursiveDataStructureTraverserTest.php
```diff
@@ -0,0 +1,147 @@
...
+		$traverser = new RecursiveDataStructureTraverser( $data );
+
+		try {
+			$traverser->insert( array( 'key' ), 'value' );
+		} catch ( \Exception $e ) {
+			$this->assertSame( 'a string', $data );
+			return;
+		}
+
+		$this->fail( 'Failed to assert that an exception was thrown when inserting a key into a string.' );
```
schlessera: For PHPUnit 4.8, the method `$this->setExpectedException( 'WP_CLI\NonExistentKeyException' );` should be used for testing exceptions.

schlessera:
@aaemnnosttv Nice work on that so far!
I've made comments across the PR. Most of it is nitpicking, but there's a couple of questions in there that we probably should discuss before copying this code over to the other entities.

Also, I'm afraid that one detail cause these commands to be dangerous, with possibly unexpected results:
The `patch <method> <key> <key-path>... <value>` signature, with an optional STDIN argument in place of the last `<value>` argument.

The problem I see here is that depending on a combination of keys and STDIN arguments, you might have the last key be accidentally used as the value instead (for example reading a file from STDIN, but that file unexpectedly turns out to be empty or causes a read error). In such a case, you might crush existing data that you intended to keep.

I'm not sure how real that danger is, but it sounds easily possible in my mind. Do you think this might be a real issue? Are there ways we could mitigate this?

src/WP_CLI/CommandWithMeta.php
danielbachhuber: > Instead of silently dying here, you should use something like the following:

I looked into this. The existing behavior is how `wp option get` operates:

```
$ wp option get fooos
$ echo $?
1
```
I think we should mirror existing behavior as much as posssible.

schlessera: Ah, yes, to avoid having the error message just be accepted as the actual value that was retrieved. Makes sense.

src/WP_CLI/RecursiveDataStructureTraverser.php
aaemnnosttv: It is used publicly within the `traverse_to` method.

src/WP_CLI/RecursiveDataStructureTraverser.php
aaemnnosttv: I wouldn't say it falls back to `patch update`, but it is the same operation with the exception that it creates any missing keys along the way. If the missing key exception is caught, it creates it and calls itself again.

phpunit.xml
aaemnnosttv: Nice catch. Must have used the wrong binary when generating the config.

features/post-meta.feature
aaemnnosttv: > What happens when you patch insert an already existing value?

Will add.

> What happens when you patch update a non-existing key? (This should produce an error.)

I believe this is covered already. See the `@patch-fail` tags.

> What happens when you patch update an existing key to the same value? (You seem to have an adapted success message in the code below.)

Will add. (Should be a notice that value did not change.)

src/WP_CLI/CommandWithMeta.php
aaemnnosttv: That part is copied from the existing `meta update` command. @danielbachhuber ?

src/WP_CLI/CommandWithMeta.php
aaemnnosttv: I believe the context is provided by the exception. Eg. `No data exists for {not-a-key}`. The message could change but I think most of it should come from the exception.

src/WP_CLI/RecursiveDataStructureTraverser.php
aaemnnosttv: Yeah, I was going to add this stuff. This has been an interactive work in progress. I didn't intend for this to be critiqued as ready for merge yet 🤠

src/WP_CLI/CommandWithMeta.php
danielbachhuber: I think the Codex is wrong, see https://github.com/wp-cli/wp-cli/issues/3190

tests/RecursiveDataStructureTraverserTest.php
aaemnnosttv: I'm aware of that utility provided by PHPUnit but I find this to be more flexible as it allows for more granular assertions rather than just expecting that exception to be thrown at some point. Doing it this way allows one to see exactly where the exception was caused and lets us make assertions after the fact. In this case, since the traverser modifies the value by reference it seems important to assert that the input data has not changed.

aaemnnosttv:
Thanks @schlessera.  I realize there are a lot of little details to clean up, most of which you highlighted but it always helps to have another set of eyes on these things.

Regarding your concern about the edge case with args & STDIN, that had crossed my mind at some point but I'm not sure there is a way to check if STDIN "isset". 

If not, then we have a few options:

- accept the risk (and document it) 
- drop support for passing the value as a positional argument, and use an associative argument. (I don't particularly like this option because it's not consistent with the other entity-value commands)
- drop support for accepting the value from STDIN (again not cool because not consistent with other commands)

As far as detecting if STDIN was intended to be used (empty or not) I haven't found an answer to.
Lack of good options has me leaning towards the first and current approach to keep both with the minor risk. Maybe there's another way?

aaemnnosttv:
Hey guys, I just pushed another update with many things addressed. One big improvement is that I found a way to reliably test if STDIN was passed or not which eliminates the possibility of the last argument being mistakenly used for the value if the STDIN is empty.

I added a method on a new `Entity\Utils` class for this although I'm guessing this could be useful for core as well. I thought we can start with it here and then just replace the import if/when it is assimilated into core?

I still need to add a few extra functional test cases that were requested, but I think we're really close to something that can be replicated to other commands.

schlessera:
Yes, let's keep the function within `Entity\Utils` for now and only refactor when it is effectively needed elsewhere.

src/WP_CLI/CommandWithMeta.php
danielbachhuber: @aaemnnosttv Bump here

src/WP_CLI/CommandWithMeta.php
danielbachhuber: @aaemnnosttv Bump here.

danielbachhuber:
@aaemnnosttv Aside from the couple remaining small nits, are you happy enough with this to land it?

@schlessera Can you give this one final review?

src/WP_CLI/Entity/RecursiveDataStructureTraverser.php
```diff
@@ -55,24 +55,49 @@ public function value() {
...
 	public function set_value( $value ) {
 		$this->data = $value;
 	}
 
-	public function delete( $locator ) {
-		$this->traverse_to( (array) $locator )->unset_on_parent();
+	/**
+	 * Unset the value at the given key path.
+	 *
+	 * @param $key_path
```
schlessera: Type hint missing. `string|int|array` ?

src/WP_CLI/Entity/RecursiveDataStructureTraverser.php
```diff
@@ -171,6 +171,7 @@ protected function create_key() {
 	 * @return bool
 	 */
 	public function exists( $key ) {
-		return ( is_array( $this->data ) && array_key_exists( $key, $this->data ) ) || ( is_object( $this->data ) && property_exists( $this->data, $key ) );
+		return ( is_array( $this->data ) && array_key_exists( $key, $this->data ) ) ||
```
schlessera: In terms of readability, pushing the operator to the next line is preferable, i.e.:
```
if ( $condition1
     || $condition2
     || $condition3 ) {
}
```

danielbachhuber:
> @aaemnnosttv Aside from the couple remaining small nits, are you happy enough with this to land it?

Bump @aaemnnosttv

aaemnnosttv:
@danielbachhuber I'm much happier with it now that it does not have the edge case that @schlessera mentioned. I'm guessing there may be a few small abstractions to make which might come up when replicating to the other commands, but I think it's in a good spot. Do you guys feel like it's functionally ready to start rolling it out to other entities?

schlessera:
@aaemnnosttv Yes, I think the functionality is good now. Feel free to attack the other commands as well, while abstracting as needed to keep the code DRY.

danielbachhuber:
@aaemnnosttv Let's go ahead and merge this then and handle the others in subsequent PRs.

danielbachhuber:
Created a new issue for `wp option (patch|pluck)` https://github.com/wp-cli/entity-command/issues/46

@aaemnnosttv As far as I can think of, `wp option *` and `wp * meta` are the only two classes of commands where this would be helpful. Can you think of any others?

aaemnnosttv:
`wp site option` and `wp transient` would benefit as well, although Transients would require a PR to the `wp-cli/cache-command` package. Thoughts?

danielbachhuber:
Good call re: `wp site option`. I don't think `wp transient` is necessary though... we can wait until someone identifies an actual need for it.

Issue #14, opened 2017-05-09
Document how to generate terms and assign to random posts


danielbachhuber:
While it probably doesn't make sense as an actual feature, it would be neat to have a code snippet for generating terms and then assigning them to random posts (e.g. to mock the situation for https://github.com/wp-cli/ideas/issues/40)

BhargavBhandari90:
Hi @danielbachhuber,
Before anything, we should decide criteria for the generating terms. I have some thought on this. 

1. What will be the names of those terms?
2. How it will assign to random posts. Like how many terms will assign to random post?

Add your thought too, if anything comes from your side.

danielbachhuber:
> 1. What will be the names ( pattern ) of those terms?

They could be the same name (e.g. `Test Term`), or same sequentially (e.g. `Test Term 1`).

> 2. How it will assign to random posts. Like how many terms will assign to random post?

We'd assign some range of terms (.e.g between 1 and 5).

BhargavBhandari90:
Hey @danielbachhuber,
Do we have a any other document like this? So that we can follow the same structure for this too.

danielbachhuber:
> Do we have a any other document like this?

Mostly the existing `## EXAMPLES` associated with each command.

Why do you include your suggested implementation here, we can refine and iterate, and then you can submit a PR?

danielbachhuber:
Punting on this https://github.com/wp-cli/entity-command/pull/362#issuecomment-1228736478

PR #30, opened 2017-07-21
Assign terms to posts

BhargavBhandari90:
For: https://github.com/wp-cli/entity-command/issues/14

### DIFF
--- a/src/Term_Command.php
+++ b/src/Term_Command.php
@@ -654,7 +654,7 @@ class Term_Command extends WP_CLI_Command {
 	 *     Assigning Terms  100% [=========] 0:02 / 0:02
 	 *
 	 */
-	public function assignment( $args, $assoc_args ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks for the pull request, @BhargavBhandari90.

The original issue was meant to be documentation, not a new command. If you'd like to create a command for this use case, I'd suggest incorporating it into the existing `wp term generate`, possibly by adding a `--post_id=<post-id>` argument that can accept a specific post id or 'random' as a value.

BhargavBhandari90:
@danielbachhuber   Actually I created that command for documentation only, because I didn't find any command which assigns terms the random post.

However, my command will assign random terms to every post.

BhargavBhandari90:
And yes, now I will try to do into `wp term generate` as you suggested @danielbachhuber.

PR #38, opened 2017-08-02
Update package with latest scaffolded components

danielbachhuber:
Related https://github.com/wp-cli/wp-cli/issues/4252


danielbachhuber:
Landing this, so we can move closer to tagging a release.

Issue #45, opened 2017-08-17
Support --order=(asc|desc) and --orderby=(default|<key>) when listing options and meta


danielbachhuber:
In `wp * meta list` and `wp option list`, we should support `--order=(asc|desc)` and `--orderby=(default|<key>)` arguments so the user has more control over the display of the results.

This might look like:

* `wp post meta list --orderby=meta_key --order=desc` - Display results ordered by `meta_key`.
* `wp option list --orderby=option_name --order=asc` - Display results ordered by `option_name`.

We'll likely need to perform these sorts in PHP because the underlying native WordPress APIs don't support these arguments.

Originally https://github.com/wp-cli/wp-cli/issues/4242

BhargavBhandari90:
@danielbachhuber 

if `--orderby=meta_key` or `--orderby=option_name` is the only option for the meta/options list then we shouldn't add that into command.
However, we can sort `options list` by its `id` too.

danielbachhuber:
> if `--orderby=meta_key` or `--orderby=option_name` is the only option for the meta/options list then we shouldn't add that into command.

There should be two options: `default` or `(meta_key|option_name)`

danielbachhuber:
🔨

PR #47, opened 2017-08-17
Add sorting option for meta list

BhargavBhandari90:
For https://github.com/wp-cli/entity-command/issues/45

### DIFF
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -40,9 +40,9 @@ Feature: List WordPress options
     And I run `wp option list --search="sample_test_field_*" --format=csv`
     Then STDOUT should be:
       """
-      option_name,option_value
-      sample_test_field_one,sample_test_field_value_one
-      sample_test_field_two,sample_test_field_value_two
+      option_id,option_name,option_value
+      1,sample_test_field_one,sample_test_field_value_one
+      2,sample_test_field_two,sample_test_field_value_two
       """
 
     When I run `wp option list --search="sample_test_field_*" --exclude="*field_one" --format=csv`
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -38,22 +38,11 @@ Feature: List WordPress options
     When I run `wp option add sample_test_field_one sample_test_field_value_one`
     And I run `wp option add sample_test_field_two sample_test_field_value_two`
     And I run `wp option list --search="sample_test_field_*" --format=csv`
-    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_name --order=asc`
-    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_value --order=asc`
     Then STDOUT should be:
       """
       option_name,option_value
       sample_test_field_one,sample_test_field_value_one
       sample_test_field_two,sample_test_field_value_two
-      """
-
-    When I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_value --order=desc`
-    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_name --order=desc`
-    Then STDOUT should be:
-      """
-      option_name,option_value
-      sample_test_field_two,sample_test_field_value_two
-      sample_test_field_one,sample_test_field_value_one
       """
 
     When I run `wp option list --search="sample_test_field_*" --exclude="*field_one" --format=csv`
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -41,15 +41,15 @@ Feature: List WordPress options
     Then STDOUT should be:
       """
       option_id,option_name,option_value
-      1,sample_test_field_one,sample_test_field_value_one
-      2,sample_test_field_two,sample_test_field_value_two
+      109,sample_test_field_one,sample_test_field_value_one
+      111,sample_test_field_two,sample_test_field_value_two
       """
 
     When I run `wp option list --search="sample_test_field_*" --exclude="*field_one" --format=csv`
     Then STDOUT should be:
       """
-      option_name,option_value
-      sample_test_field_two,sample_test_field_value_two
+      option_id,option_name,option_value
+      111,sample_test_field_two,sample_test_field_value_two
       """
 
     When I run `wp option list`
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -45,6 +45,16 @@ Feature: List WordPress options
       111,sample_test_field_two,sample_test_field_value_two
       """
 
+    When I run `wp option list --search="sample_test_field_*" --orderby=option_name --order=desc --format=csv`
+    And I run `wp option list --search="sample_test_field_*" --orderby=option_value --order=desc --format=csv`
+    And I run `wp option list --search="sample_test_field_*" --orderby=option_id --order=desc --format=csv`
+    Then STDOUT should be:
+      """
+      option_id,option_name,option_value
+      111,sample_test_field_two,sample_test_field_value_two
+      109,sample_test_field_one,sample_test_field_value_one
+      """
+
     When I run `wp option list --search="sample_test_field_*" --exclude="*field_one" --format=csv`
     Then STDOUT should be:
       """
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -38,28 +38,29 @@ Feature: List WordPress options
     When I run `wp option add sample_test_field_one sample_test_field_value_one`
     And I run `wp option add sample_test_field_two sample_test_field_value_two`
     And I run `wp option list --search="sample_test_field_*" --format=csv`
+    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_name --order=asc`
+    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_value --order=asc`
     Then STDOUT should be:
       """
-      option_id,option_name,option_value
-      109,sample_test_field_one,sample_test_field_value_one
-      111,sample_test_field_two,sample_test_field_value_two
+      option_name,option_value
+      sample_test_field_one,sample_test_field_value_one
+      sample_test_field_two,sample_test_field_value_two
       """
 
-    When I run `wp option list --search="sample_test_field_*" --orderby=option_name --order=desc --format=csv`
-    And I run `wp option list --search="sample_test_field_*" --orderby=option_value --order=desc --format=csv`
-    And I run `wp option list --search="sample_test_field_*" --orderby=option_id --order=desc --format=csv`
+    When I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_value --order=desc`
+    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_name --order=desc`
     Then STDOUT should be:
       """
-      option_id,option_name,option_value
-      111,sample_test_field_two,sample_test_field_value_two
-      109,sample_test_field_one,sample_test_field_value_one
+      option_name,option_value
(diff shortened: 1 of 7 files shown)


miya0001:
Please add tests too. 😄

BhargavBhandari90:
@miya0001   I am little confused here about `--orderby` for `meta list`. By which fields we can sort the `meta list`?. As of now it's possible by `metakey` only.

However, I am improving the code for now. Let me know if I am doing anything wrong.

miya0001:
@BhargavBhandari90 

Isn't there `default`?
https://github.com/wp-cli/entity-command/issues/45#issuecomment-323197093

miya0001:
Ah, I understood that `meta list` doesn't have any params to sort the results.
https://github.com/wp-cli/entity-command/blob/master/src/WP_CLI/CommandWithMeta.php#L44

BhargavBhandari90:
@miya0001 
I am not sure why my tests are failing for 

102.4
PHP: 5.6
WP_VERSION=3.7.11

AND

102.5
PHP: 5.6
WP_VERSION=trunk

Do you have any idea?

schlessera:
@BhargavBhandari90 You cannot just rely on a fixed option ID, because the number of default options can change from one WP version to the next. This is why your tests are failing, the number of options is different, producing different option IDs.

src/Option_Command.php
```diff
@@ -218,9 +237,10 @@ public function list_( $args, $assoc_args ) {
 		global $wpdb;
 		$pattern = '%';
 		$exclude = '';
-		$fields = array( 'option_name', 'option_value' );
+		$fields = array( 'option_id', 'option_name', 'option_value' );
```
schlessera: You should just add the ID here, that is a breaking change. People who are already using this command will not expect the format to suddenly change from one version ot the next.

If you need the ID for your tests, you can adapt the tests to specifically add the `option_id` to the list of displayed fields instead.

src/Option_Command.php
```diff
@@ -273,8 +293,16 @@ public function list_( $args, $assoc_args ) {
...
+		// Ser orderby field.
+		$orderby_field = 'option_name';
+		if ( 'option_value' === \WP_CLI\Utils\get_flag_value( $assoc_args, 'orderby' ) ) {
+			$orderby_field = 'option_value';
+		} else if ( 'option_id' === \WP_CLI\Utils\get_flag_value( $assoc_args, 'orderby' ) ) {
+			$orderby_field = 'option_id';
+		}
+
+		$results = $wpdb->get_results( "SELECT `option_id`,`option_name`,`option_value`,`autoload`" . $size_query
+					. " FROM `$wpdb->options` {$where} ORDER BY `$wpdb->options`.`$orderby_field` $sort" );
```
schlessera: The `order_by` should be optional, so that when you don't specify it, you get the results in the order that they are stored in the DB. This is better for performance and keeps the current existing behavior intact.

src/Option_Command.php
```diff
@@ -170,6 +170,24 @@ public function add( $args, $assoc_args ) {
 	 *   - total_bytes
 	 * ---
 	 *
+	 * [--orderby=<fields>]
+	 * : Set orderby which field.
+	 * ---
+	 * default: option_name
```
danielbachhuber: The default orderby behavior is `option_id`, I believe.

src/WP_CLI/CommandWithMeta.php
```diff
@@ -31,40 +31,68 @@
...
+		// Get metadata.
+		if ( 'user' === $this->meta_type ) {
+			// Where clause.
+			$where    = $wpdb->prepare( " WHERE `$wpdb->usermeta`.`user_id` = %d ", $object_id );
+			$metadata = $wpdb->get_results( "SELECT `user_id`,`meta_key`,`meta_value` FROM `$wpdb->usermeta` {$where} ORDER BY `$wpdb->usermeta`.`$orderby_field` $sort" );
+		} else {
+			// Where clause.
+			$where    = $wpdb->prepare( " WHERE `$wpdb->postmeta`.`post_id` = %d ", $object_id );
+			$metadata = $wpdb->get_results( "SELECT `post_id`,`meta_key`,`meta_value` FROM `$wpdb->postmeta` {$where} ORDER BY `$wpdb->postmeta`.`$orderby_field` $sort" );
+		}
```
danielbachhuber: Per my original comment, can we perform the orderby with PHP instead of SQL, so we can continue to use `get_metadata()`?

src/Option_Command.php
BhargavBhandari90: I checked it. It is sorting by `option_name` in ascending order.

features/option-list.feature
```diff
@@ -38,13 +38,24 @@ Feature: List WordPress options
     When I run `wp option add sample_test_field_one sample_test_field_value_one`
     And I run `wp option add sample_test_field_two sample_test_field_value_two`
     And I run `wp option list --search="sample_test_field_*" --format=csv`
+    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_name --order=asc`
+    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_value --order=asc`
```
danielbachhuber: Why do you run these commands twice?

features/user-meta.feature
```diff
@@ -61,7 +61,7 @@ Feature: Manage user custom fields
     When I run `wp user meta list 1 --format=json --keys=nickname,foo --fields=meta_key,meta_value`
     Then STDOUT should be JSON containing:
       """
-      [{"meta_key":"nickname","meta_value":"admin"},{"meta_key":"foo","meta_value":["1","2"]}]
+      [{"meta_key":"foo","meta_value":["1","2"]},{"meta_key":"nickname","meta_value":"admin"}]
```
danielbachhuber: Given the default sort order shouldn't be changing, this test shouldn't be changing either.

danielbachhuber:
Looks like there's still a few things to clean up.

schlessera:
@danielbachhuber Do you know how to deal with existing change requests that were addressed? I had now accepted the previous change requests I made, but this looks like a general green-light for the entire PR. GitHub does not really seem to make any distinction there...

features/option-list.feature
schlessera: The ordering is different at the end. This produces two different lines of output, that are tested below.

schlessera: But the general structure is not really logical to me...

danielbachhuber:
> Do you know how to deal with existing change requests that were addressed? I had now accepted the previous change requests I made, but this looks like a general green-light for the entire PR. GitHub does not really seem to make any distinction there...

I think the code review workflow focuses on the pull request, not individual changes.

features/option-list.feature
BhargavBhandari90: @danielbachhuber all three commands have the same output. That's why I added like this.

Should I make it separate?

miya0001: You can do that like following.
http://docs.behat.org/en/v2.5/guides/1.gherkin.html

```
    When I run `wp option add sample_test_field_one sample_test_field_value_one`
    And I run `wp option add sample_test_field_two sample_test_field_value_two`
    And I run `wp option list --search="sample_test_field_*" --format=csv`
    Then STDOUT should be:
      """
      option_name,option_value
      sample_test_field_one,sample_test_field_value_one
      sample_test_field_two,sample_test_field_value_two
      """

    When I run `wp option add sample_test_field_one sample_test_field_value_one`
    And I run `wp option add sample_test_field_two sample_test_field_value_two`
    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_name --order=asc`
    Then STDOUT should be:
      """
      option_name,option_value
      sample_test_field_one,sample_test_field_value_one
      sample_test_field_two,sample_test_field_value_two
      """

    When I run `wp option add sample_test_field_one sample_test_field_value_one`
    And I run `wp option add sample_test_field_two sample_test_field_value_two`
    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_value --order=asc`
    Then STDOUT should be:
      """
      option_name,option_value
      sample_test_field_one,sample_test_field_value_one
      sample_test_field_two,sample_test_field_value_two
      """
```

src/WP_CLI/CommandWithMeta.php
```diff
@@ -79,13 +80,50 @@ public function list_( $args, $assoc_args ) {
...
+				);
 			}
 
 		}
 
-		print_r( $items );
+		$order   = Utils\get_flag_value( $assoc_args, 'order' );
+		$orderby = Utils\get_flag_value( $assoc_args, 'orderby' );
+
+		// Sort array.
```
schlessera: The sorting algorithm should be refactored to DRY it up. Instead of 4 functions, you should only use 1 function.

Investigate: 
```
usort( $items, function ( $a, $b ) use ( $orderby, $order ) {
    // $orderby and $order available in here as well now.
    // You can now use something like `$a->$orderby`.
}
```

BhargavBhandari90: Ohh thanks @schlessera  I was searching for this kind of logic.

schlessera:
@BhargavBhandari90 I don't fully understand your latest change. You're still not using the `$orderby` and `$order` variables to make the logic DRY.

It should be something like this:
```
if ( 'id' !== $orderby ) {
	usort( $items, function ( $a, $b ) use ( $orderby, $order ) {
		return 'asc' === $order
			? strcmp( $a->$orderby, $b->$orderby )
			: strcmp( $b->$orderby, $a->$orderby );
	});
} elseif ( 'id' === $orderby && 'desc' === $order ) { // Sort by default descending.
	krsort( $items );
}
```

src/Option_Command.php
```diff
@@ -273,8 +292,17 @@ public function list_( $args, $assoc_args ) {
 		}
 		$where .= $autoload_query . $transients_query;
 
+		// Set orderby field.
+		$orderby = '';
+		if ( 'option_name' === \WP_CLI\Utils\get_flag_value( $assoc_args, 'orderby' ) && 'asc' === $sort ) {
```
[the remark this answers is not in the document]
BhargavBhandari90: Now I used `usort` for this.

src/Option_Command.php
BhargavBhandari90: Yes, you are right. Made some changes according to that.

features/option-list.feature
BhargavBhandari90: Thanks @miya0001

BhargavBhandari90:
@danielbachhuber  @schlessera  and @miya0001 Please review this.

features/option-list.feature
```diff
@@ -64,3 +64,56 @@ Feature: List WordPress options
       sample_test_field_one
       """
 
+  Scenario: List option with sorting option
+    Given a WP install
+
+    When I run `wp option add sample_test_field_one sample_test_field_value_one`
+    And I run `wp option add sample_test_field_two sample_test_field_value_two`
+    And I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_id --order=asc`
```
miya0001: Please change line 68-72 like following, so it will be easier to read.

```
    Given a WP install
    And I run `wp option add sample_test_field_one sample_test_field_value_one`
    And I run `wp option add sample_test_field_two sample_test_field_value_two`

    When I run `wp option list --search="sample_test_field_*" --format=csv --orderby=option_id --order=asc`
```

BhargavBhandari90: Got your point @miya0001

danielbachhuber:
@schlessera Can you give this a final review/merge?

PR #53, opened 2017-08-21
update document in user_command.php file

### LINKED ISSUE #52: Need document update in User_Command.php file
Need document update in User_Command.php file

Umangvaghela:
Fix #52


danielbachhuber:
@Umangvaghela Could you explain in greater detail what you're trying to do here?

Umangvaghela:
@danielbachhuber
why we are user and <role> and [<role>].Can we use one of them to maintain the consistency?

PR #57, opened 2017-08-22
Update coding standards for condition

### LINKED ISSUE #56: Condition is not  proper in Site_Command.php create()
Use proper condition.

Umangvaghela:
fix #56

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -315,7 +315,7 @@ class Site_Command extends \WP_CLI\CommandWithDBObject {
 		// Network
 		if ( !empty( $assoc_args['network_id'] ) ) {
 			$network = $this->_get_network( $assoc_args['network_id'] );
-			if ( $network === false ) {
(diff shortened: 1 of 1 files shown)


src/Site_Command.php
```diff
@@ -369,7 +369,7 @@ public function create( $_, $assoc_args ) {
 		if ( !$user_id ) { // Create a new user with a random password
 			$password = wp_generate_password( 12, false );
 			$user_id = wpmu_create_user( $base, $password, $email );
-			if ( false == $user_id ) {
+			if ( false === $user_id ) {
```
miya0001: `$user_id` can be `0`, please restore this line.

miya0001: Please see following:
https://developer.wordpress.org/reference/functions/wpmu_create_user/

Umangvaghela:
@miya0001 

Thanks for the approved change request.

PR #55, opened 2017-08-22
Remove Unused Parameter in function

Umangvaghela:
fix #54 issue

PR #58, opened 2017-08-22
remove unwanted parameter

Umangvaghela:
fix #55 issue


miya0001:
Hi @Umangvaghela 

Please don't send same PR.
See https://github.com/wp-cli/entity-command/pull/55#issuecomment-323973598.

Thanks!

PR #59, opened 2017-08-22
set proper condition with use of strict comparison

Umangvaghela:
update condition with use of strict comparison


danielbachhuber:
> set proper condition with use of strict comparison

Because WordPress isn't strict about types internally, this change could have unintended consequences.

@Umangvaghela Please join us in the `#cli` channel in Slack to discuss future pull requests before submitting them.

PR #62, opened 2017-08-24
add correct date formate

Umangvaghela:
add correct date formate

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -424,7 +424,7 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 	 * default:
 	 * ---
 	 *
-	 * [--post_date=<yyyy-mm-dd hh:ii:ss>]
+	 * [--post_date=<yyyy-mm-dd-hh-ii-ss>]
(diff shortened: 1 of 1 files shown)


src/Post_Command.php
```diff
@@ -424,7 +424,7 @@ public function list_( $_, $assoc_args ) {
 	 * default:
 	 * ---
 	 *
-	 * [--post_date=<yyyy-mm-dd>]
+	 * [--post_date=<yyyy-mm-dd hh:ii:ss>]
```
danielbachhuber: This should actually be `<yyyy-mm-dd-hh-ii-ss>`, see https://github.com/wp-cli/entity-command/pull/43

miya0001:
@danielbachhuber 

> This should actually be <yyyy-mm-dd-hh-ii-ss>, see #43

In this case, we can pass the option in the following 2 patterns? 

* `--post_date=2018-01-01-00-00-00`
* `--post_date='2018-01-01 00-00-00'`

Should we update examples?

danielbachhuber:
> Should we update examples?

Sure, feel free to do so in a follow-up PR.

Issue #51, opened 2017-08-20
Introduce wp site mature and wp site public


johnbillion:
Commands for marking a site as active/deactivated, archived/unarchived, spam/unspammed already exist. These change the site status via `update_blog_status()`.

Marking a site as mature/unmature and public/unpublic are missing (the terminology here is terrible). The following commands should be added for completeness:

* `wp site mature <id>`
* `wp site public <id>`

Sidsector9:
@johnbillion, can we discuss about the terminologies? 

How does public/private and mature/(all|general) sound?

Sidsector9:
#63

danielbachhuber:
🔨

PR #63, opened 2017-08-24
Added support for mature and public status types

Sidsector9:
See #51

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -652,7 +652,7 @@ class Site_Command extends \WP_CLI\CommandWithDBObject {
 	}
 
 	/**
-	 * Set one or more sites to mature.
+	 * Set one or more sites as mature.
 	 *
 	 * ## OPTIONS
 	 *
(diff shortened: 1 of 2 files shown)


miya0001:
Related #51

miya0001:
Can you add behat tests?

Sidsector9:
@miya0001 almost done, will push the tests soon

src/Site_Command.php
```diff
@@ -651,6 +651,82 @@ public function unspam( $args ) {
 		$this->update_site_status( $args, 'spam', 0 );
 	}
 
+	/**
+	 * Set one or more sites to mature.
```
danielbachhuber: Set one or more sites **as** mature

src/Site_Command.php
```diff
@@ -651,6 +651,82 @@ public function unspam( $args ) {
...
+	 *
+	 * <id>...
+	 * : One or more IDs of sites to set as mature.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     $ wp site mature 123
+	 *     Success: Site 123 set to mature.
+	 *
+	 * @subcommand mature
```
danielbachhuber: This `@subcommand mature` declaration isn't necessary when it mirrors the name of the method. `@subcommand` is only used when the command name differs from the method name.

src/Site_Command.php
```diff
@@ -651,6 +651,82 @@ public function unspam( $args ) {
...
+	 * : One or more IDs of sites to set for general audience.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     $ wp site general 123
+	 *     Success: Site 123 marked for general audience.
+	 *
+	 * @subcommand general
+	 */
+	public function general( $args ) {
```
danielbachhuber: For parity with the other commands, let's call this `unmature` instead of `general`. As it stands, it's not clear that `general` is the opposite of `mature`.

danielbachhuber:
@Sidsector9 As a point of procedure, if you include the issue in the body of the pull request (see screenshot) then the pull request will correctly appear in the history of the issue:

image

image

Sidsector9:
@danielbachhuber The changes are done

PR #73, opened 2017-08-31
Error and prevent deleting root site

### LINKED ISSUE #66: wp site delete -- idiot-proof against deleting root blog
I'd like to suggest setting a test for which blog to delete, so that you don't accidentally delete the root blog the way I did! I had a function, feeding a value into site-delete, with the plan to delete several hundred blogs at a time (passing the --yes flag to the command). 

But I  got distracted and didn't notice what the return value was going to be. It was "1".  

Maybe it shouldn't be possible to delete the blog that matches BLOG_ID_CURRENT_SITE.

Sidsector9:
Fixes #66

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -265,7 +265,12 @@ class Site_Command extends \WP_CLI\CommandWithDBObject {
 
 		$site_url = trailingslashit( $blog->siteurl );
 
-		WP_CLI::confirm( "Are you sure you want to delete the '$site_url' site?", $assoc_args );
+		if ( $blog_id == BLOG_ID_CURRENT_SITE && ! \WP_CLI\Utils\get_flag_value( $assoc_args, 'delete-root' ) ) {
+			WP_CLI::line( WP_CLI::colorize( "%R'$site_url' is the root site. Pass the --delete-root flag to delete it.%n" ), $assoc_args );
+			return;
+		} else {
+			WP_CLI::confirm( "Are you sure you want to delete the '$site_url' site?", $assoc_args );
+		}
 
 		wpmu_delete_blog( $blog->blog_id, ! \WP_CLI\Utils\get_flag_value( $assoc_args, 'keep-tables' ) );
 
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -257,7 +257,7 @@ class Site_Command extends \WP_CLI\CommandWithDBObject {
 			$blog_id = $args[0];
 
 			if ( is_main_site( $blog_id ) ) {
-				WP_CLI::error( 'You can\'t delete root blog.' );
+				WP_CLI::error( 'You cannot delete the root site.' );
 			}
 
 			$blog = get_blog_details( $blog_id );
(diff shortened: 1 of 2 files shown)


src/Site_Command.php
```diff
@@ -265,7 +265,12 @@ function delete( $args, $assoc_args ) {
 
 		$site_url = trailingslashit( $blog->siteurl );
 
-		WP_CLI::confirm( "Are you sure you want to delete the '$site_url' site?", $assoc_args );
+		if ( $blog_id == BLOG_ID_CURRENT_SITE && ! \WP_CLI\Utils\get_flag_value( $assoc_args, 'delete-root' ) ) {
+			WP_CLI::line( WP_CLI::colorize( "%R'$site_url' is the root site. Pass the --delete-root flag to delete it.%n" ), $assoc_args );
+			return;
+		} else {
+			WP_CLI::confirm( "Are you sure you want to delete the '$site_url' site?", $assoc_args );
+		}
```
danielbachhuber: Given we already have a confirmation prompt before deleting a site, I don't think it's necessary to require another flag to delete the root site.

Instead, when deleting the root site we should customize the confirmation message to indicate what happens when you delete the root site.

As currently implemented, this would be a breaking change

Sidsector9: For a scenario as:

```
http://wpcli.com
http://one.wpcli.com
http://two.wpcli.com
```
Following observations have been made after executing
`wp site delete 1`

1. The tables aren't dropped but the users from `wpcli.com` blog are deleted.
2. `wpcli.com` gets deactivated ( You can only activate it again through WP-CLI since the admin dashboard doesn't give you an option to activate/deactivate the main site. )

Irrelevant of the fact that whether you want to keep/drop tables, the users will get deleted as evident [here](https://core.trac.wordpress.org/browser/tags/4.8/src/wp-admin/includes/ms.php#L82)

src/Site_Command.php
```diff
@@ -265,7 +265,12 @@ function delete( $args, $assoc_args ) {
 
 		$site_url = trailingslashit( $blog->siteurl );
 
-		WP_CLI::confirm( "Are you sure you want to delete the '$site_url' site?", $assoc_args );
+		if ( $blog_id == BLOG_ID_CURRENT_SITE && ! \WP_CLI\Utils\get_flag_value( $assoc_args, 'delete-root' ) ) {
```
spacedmonkey: Use is_main_site here

Sidsector9:
Related to https://github.com/wp-cli/entity-command/issues/66#issuecomment-327319456 and https://github.com/wp-cli/entity-command/issues/66#issuecomment-327437081
I have made the necessary modifications.

src/Site_Command.php
```diff
@@ -251,16 +251,20 @@ function delete( $args, $assoc_args ) {
...
 		} else {
 			if ( empty( $args ) ) {
-				WP_CLI::error( "Need to specify a blog id." );
+				WP_CLI::error( 'Need to specify a blog id.' );
 			}
 
 			$blog_id = $args[0];
 
+			if ( is_main_site( $blog_id ) ) {
+				WP_CLI::error( 'You can\'t delete root blog.' );
```
danielbachhuber: Should we explain why the root blog can't be deleted?

Sidsector9: I don't think that's necessary, users will do understand that WordPress doesn't allow deletion of root blog.

danielbachhuber: Ok. Updated the language a bit in abae4141

Issue #50, opened 2017-08-20
Introduce wp user spam


johnbillion:
The `wp site spam` command already exists for spamming sites (on a Multisite network), but there's no corresponding command for spamming users. `wp user spam` should be added.

PR #74, opened 2017-09-02
Add `wp user spam` command

BhargavBhandari90:
Command:
- wp user spam 123

For : https://github.com/wp-cli/entity-command/issues/50

### DIFF
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -1034,19 +1034,24 @@ class User_Command extends \WP_CLI\CommandWithDBObject {
 		global $wpdb;
 		$user_ids = $args;
 
+		// If site is not multisite, then stop execution.
+		if ( ! is_multisite() ) {
+			WP_CLI::error( 'Sorry! this command is for multisite only.' );
+		}
+
 		foreach ( $user_ids as $user_id ) {
 
 			$user = get_userdata( $user_id );
 
 			// If no user found, then show warning.
 			if ( empty( $user ) ) {
-				WP_CLI::warning( sprintf( __( 'User %d doesn\'t exist.' ), esc_html( $user_id ) ) );
+				WP_CLI::warning( sprintf( 'User %d doesn\'t exist.', esc_html( $user_id ) ) );
 				continue;
 			}
 
 			// Super admin should not be marked as spam.
 			if ( is_super_admin( $user->ID ) ) {
-				WP_CLI::warning( sprintf( __( 'User cannot be modified. The user %d is a network administrator.' ), esc_html( $user->ID ) ) );
+				WP_CLI::warning( sprintf( 'User cannot be modified. The user %d is a network administrator.', esc_html( $user->ID ) ) );
 				continue;
 			}
 
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -1,5 +1,7 @@
 <?php
 
+use \WP_CLI\Utils;
+
 /**
  * Manage users.
  *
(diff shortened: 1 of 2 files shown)


miya0001:
I think we shouldn't try to use SQL as much as possible.
Can you use `update_user_status()` or so?

BhargavBhandari90:
yes, @miya0001 I just saw that function. Applying that only.

miya0001:
Thanks @BhargavBhandari90 😄

BhargavBhandari90:
Behat test for user.feature shows following error for `spam` command

> PHP Fatal error:  Call to undefined function get_network() in src/User_Command.php on line 1064
      Fatal error: Call to undefined function get_network() in src/User_Command.php on line 1064
      cwd: /tmp/wp-cli-test-run-user-test.feature.3-59abbe9ab397d9.99056625/
      exit status: 255

However, I am not getting this error while running command into the terminal.

So I am confused, is my code has something wrong or what?

miya0001:
Ah, I guess we should do like following.
https://github.com/wp-cli/entity-command/blob/master/features/site.feature#L4

BhargavBhandari90:
@miya0001 Yes, see I just did that. That was real time :stuck_out_tongue:

miya0001:
Hi,

If `wp user spma` is passed multiple arguments and there is an error from one of them, then status code should be `1`. 
But I guess your implementation will return `0`.

We need consistency between same commands, so following is a good example to handle multiple arguments.

* https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L258-L293
* https://github.com/wp-cli/extension-command/blob/master/features/plugin-activate.feature#L15-L35

BhargavBhandari90:
ok. Checking this.

BhargavBhandari90:
One test is failing. because I used a function which was not available in that WP version.
How to tackle that situation?

Here is the error https://travis-ci.org/wp-cli/entity-command/jobs/271378052

miya0001:
Do you want to use `get_network()` to mark spam the user's site too?
I guess there is a good example. :)
https://github.com/wp-cli/entity-command/blob/master/src/Site_Command.php#L749-L766

BhargavBhandari90:
> If wp user spam is passed multiple arguments and there is an error from one of them, then status code should be 1.
But I guess your implementation will return 0.

@miya0001 As you said we need `wp user unspam` for this. So I will implement this.

BhargavBhandari90:
Hi @miya0001 

> If wp user spam is passed multiple arguments and there is an error from one of them, then status code should be 1.

I have implemented this thing. But I don't know how to get retun code so that I can add it into feature file.

Currently I am not adding that line into `users.feature`.

miya0001:
@BhargavBhandari90 

Please write test like following.

```
When I try `wp user spam {BBEE_ID} 9999`
Then STERR should contain:
  """
  Warning: User 9999 doesn't exist.
  """
And the return code should be 1
```

miya0001:
Ah, sorry, it should be like following.

```
When I try `wp user spam {BBEE_ID} 9999`
Then STOUT should contain:
  """
  User {BBEE_ID} marked as spam.
  """
Then STOUT should contain:
  """
  Warning: User 9999 doesn't exist.
  """
And the return code should be 1
```

BhargavBhandari90:
@miya0001 This should be fine now.

src/User_Command.php
```diff
@@ -1017,4 +1020,107 @@ private static function wp_new_user_notification( $user_id, $password ) {
...
+	}
+
+	/**
+	 * Common command for updating user data.
+	 */
+	private function update_msuser_status( $user_ids, $pref, $value ) {
+
+		// If site is not multisite, then stop execution.
+		if ( ! is_multisite() ) {
+			WP_CLI::error( 'Sorry! this command is for multisite only.' );
```
schlessera: This should mimic the other commands:
```
WP_CLI::error( 'This is not a multisite install.' );
```

schlessera:
The grammar is of for the batch results, but this needs to be fixed in `WP_CLI\Utils`. I've opened an issue for this here: https://github.com/wp-cli/wp-cli/issues/4338

src/User_Command.php
BhargavBhandari90: Done. @schlessera

schlessera:
Thanks, @BhargavBhandari90 !

PR #79, opened 2017-09-05
Yoda condition

mitraval192:
I have added the yoda condition on few condition statements


miya0001:
Thanks PR, but it isn't woking for us as expected.

Issue #84, opened 2017-09-07
wp user session commands use incorrect terminology


johnbillion:
The `wp user session` commands use the word "token" in several places. This is incorrect, as the data being described is actually the "verifier", which is different from the token.

The session token identifies the session, but is not persistently stored. It's only stored in a user's auth and logged_in cookies. `wp user session` does not deal with the token, but the verifier, which is a one-way hash of the token, and which is stored persistently.

If a user tries to interact with the `wp user session` command using a token instead of a verifier, they're going to have a hard time and may wonder why nothing works, especially if they cross-reference things with the `WP_Session_Tokens` class in core.

miya0001:
I agree but we have an option like following.

```
$ wp session list --fields=token
```

If we change the `token` to `verifier`, it will be breaking change. 😓

johnbillion:
> If we change the token to verifier, it will be breaking change. 😓

Yep. I'm not sure how best to address this. See #85.

schlessera:
So the `verifier` is a hash of the `token`. Would it be "good enough" to add something like this to the command's documentation?

```
Note: The `token` field does not return the actual token, but a hash of it. The real token is not persisted and can only be found in the corresponding cookies on the client side.
```

johnbillion:
Yeah that sounds like a better approach than introducing a breaking change.

PR #85, opened 2017-09-07
Correct the terminology for user session verifiers vs tokens

johnbillion:
See #84.

This corrects the terminology, but it technically a breaking a change as the return value of `wp user session list` will change. Feedback welcome on how to address this.


johnbillion:
Closing in favour of adding some help docs instead.

PR #95, opened 2017-10-01
Add definition for arguments supported by wp_insert_post() and wp_update_post()

### LINKED ISSUE #44: Define known fields as allowed arguments for `wp post (create|update)`
Same thing as #19, but for `wp post create` and `wp post update`.

Under the hood, both commands use `wp_insert_post()`. We should define `wp_insert_post()`'s arguments as allowed arguments for each command. `[--<field>=<value>]` should remain as an argument to accept additional arbitrary values.

emgk:
This PR fixes #44 

I have added the definitions of the each args ( but not 'post_category', 'tax_input', and 'meta_input' as these field needs array values ) which passed to wp_insert_post or wp_update_post. 

\cc @danielbachhuber @schlessera

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -46,67 +46,67 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 	 *   Passing `-` as the filename will cause post content to
 	 *   be read from STDIN.
 	 *
-	 * [--<post_author>=<post_author>]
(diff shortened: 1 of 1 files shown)


schlessera:
@emgk You've used the wrong syntax for these, which makes the etsts fail.

Instead of something like `[--<post_type>=<post_type>]`, you need to use `--post_type=<post_type>`.

PR #100, opened 2017-10-07
`wp user import-csv` to be able to import CSV from STDIN

### LINKED ISSUE #96: `wp user import-csv` to be able to import CSV from STDIN
Almost all commands which have a positional argument `<file>` can import data from STDIN.
I think `wp user import-csv` should support it too.

For example:
```
$ cut -d ',' -f 1,3,4,8 new_users.2017-10-02.csv | wp user import-csv -
```

miya0001:
Fixes #96

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 vendor/
 *.zip
 *.tar.gz
+composer.lock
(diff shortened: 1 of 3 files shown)


gitlost:
A small thing I noticed is that the second `has_stdin()` check should be removed from the `if ( '-' === $filename ` statement before the loop.

PR #105, opened 2017-10-29
Add a "network" command namespace.

schlessera:
This allows the addition of a description to show up in the help screen.

See https://github.com/wp-cli/wp-cli/issues/4385


schlessera:
This depends on https://github.com/wp-cli/wp-cli/pull/4470 .

PR #117, opened 2017-12-17
Replace use of "install" with "installation"

thrijith:
here is the link to the [issue](https://github.com/wp-cli/wp-cli/issues/4541)
Replaced use of "install" with "installation" in `src/Site_Command.php`


thrijith:
@danielbachhuber thanks for the heads up, will keep this in mind for next time.

PR #119, opened 2017-12-18
Add a new command for resetting a user password

danielbachhuber:


src/User_Command.php
```diff
@@ -1013,6 +1013,36 @@ public function import_csv( $args, $assoc_args ) {
...
+		$fetcher = new \WP_CLI\Fetchers\User;
+		$users = $fetcher->get_many( $args );
+		foreach( $users as $user ) {
+			wp_update_user( array( 'ID' => $user->ID, 'user_pass' => wp_generate_password() ) );
+			WP_CLI::log( "Reset password for {$user->user_login}." );
+		}
+		if ( $skip_email ) {
+			remove_filter( 'send_password_change_email', '__return_false' );
+		}
+		WP_CLI::success( 'Passwords reset.' );
```
swissspidy: I don't know how other commands handle this, but can we change the success message a bit? Like this:

``` php
if ( 1 === count( $users ) ) {
	WP_CLI::success( 'Password reset.' );
} else {
	WP_CLI::success( 'Passwords reset.' );
}
```

danielbachhuber: > I don't know how other commands handle this, but can we change the success message a bit?

Good catch. The other commands pluralize based on the number of entities affected.

Updated in 7a575ba7c44a91329b973c3dab058ed5a0de8210

Issue #121, opened 2017-12-21
Consider defaulting to `--no-transients` when listing options


danielbachhuber:
When I run `wp option list`, I _always_ mean to list available options and I _never_ mean to also see all data stored in transients.

Although they use the same underlying storage mechanism, they're conceptually different APIs in WordPress. 

It'd be nice if `wp option list` defaulted to not displaying transients unless I explicitly passed the `--transients` flag.

PR #122, opened 2017-12-30
Exclude transient from list

BhargavBhandari90:
For: https://github.com/wp-cli/entity-command/issues/121

### DIFF
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -57,6 +57,10 @@ Feature: List WordPress options
       """
       sample_test_field_one
       """
+    And STDOUT should not contain:
+      """
+      _transient
+      """
 
     When I run `wp option list --exclude="sample_test_field_one"`
     Then STDOUT should not contain:
--- a/features/option-list.feature
+++ b/features/option-list.feature
@@ -121,3 +121,21 @@ Feature: List WordPress options
       sample_test_field_one,sample_test_field_value_one
       """
 
+  Scenario: Default list option without transient
+    Given a WP install
+    And I run `wp transient set wp_transient_flag wp_transient_flag`
+
+    When I run `wp option list`
+    Then STDOUT should not contain:
+      """
+      wp_transient_flag
+      """
+    And STDOUT should not contain:
+      """
+      _transient
+      """
+    And STDOUT should contain:
+      """
+      siteurl
+      """
+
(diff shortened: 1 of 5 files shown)


schlessera:
Thanks for the PR, @BhargavBhandari90 !

Can you add another test that makes use of the flag you've added, to make sure transients are shown in that case?

BhargavBhandari90:
@schlessera The test for `transient` flag is already available at here https://github.com/wp-cli/entity-command/blob/master/features/option-list.feature#L21

BhargavBhandari90:
Hey @schlessera it's showing `  The "https://getcomposer.org/versions" file could not be downloaded: failed to open stream: Connection timed out  
`  at here. https://travis-ci.org/wp-cli/entity-command/jobs/324292393

I think build needs to be restarted again. Not sure about hat

gitlost:
Restarted that for you...

BhargavBhandari90:
Thanks @gitlost . It worked.

BhargavBhandari90:
@schlessera Added a test.

schlessera:
@BhargavBhandari90 I'm sorry, I only realized just now that you have used the wrong mechanism to include the default transient.

You shouldn't set the `$exclude` variable, because that will then override any `--exclude=<something>` parameter that a user might have provided.

Instead, you should change the logic in https://github.com/wp-cli/entity-command/blob/master/src/Option_Command.php#L280-L287 and provide a default value of `false` instead of `null`:

```php
$show_transients = Utils\get_flag_value( $assoc_args, 'transients', false );
if ( $show_transients ) {
	$transients_query = " AND option_name LIKE '\_transient\_%'
	OR option_name LIKE '\_site\_transient\_%'";
} else {
	$transients_query = " AND option_name NOT LIKE '\_transient\_%'
	AND option_name NOT LIKE '\_site\_transient\_%'";
}
```

BhargavBhandari90:
@schlessera Oh !!! Got it. Will make the change in this weekend. Thanks.

BhargavBhandari90:
Hey @gitlost again it's time out https://travis-ci.org/wp-cli/entity-command/builds/325780058?utm_source=github_status&utm_medium=notification

BhargavBhandari90:
Sending new PR. I merged `master` and got other changes in my PR.

PR #124, opened 2018-01-05
Use correct verb in example for deleting trashed posts

ocean90:
It's "Deleted", not "Trashed".


schlessera:
Thanks for the PR. @ocean90 !

PR #127, opened 2018-01-06
Remove transient from option list

BhargavBhandari90:
For #121


BhargavBhandari90:
@schlessera Please consider this new PR. The test is failing due to timeout. I think this needs to be restarted.

schlessera:
Great, this should be the correct fix now.

Thanks for your persistence, @BhargavBhandari90 !

BhargavBhandari90:
Thanks for you help @schlessera

PR #129, opened 2018-01-14
update post create/update

thrijith:
to fix wp-cli/wp-cli#4605
update the condition to set post category if name/slug  is given in `--post_category`
work in progress

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -29,6 +29,28 @@ Feature: Manage WordPress posts
     When I try the previous command again
     Then the return code should be 1
 
+    When I run `wp term create category "First Category" --porcelain`
+    And save STDOUT as {TERM_ID}
+    And I run `wp term create category "Second Category" --porcelain`
+    And save STDOUT as {SECOND_TERM_ID}
+
+    When I run `wp post create --post_title="Test category" --post_category="First Category" --porcelain`
+    Then STDOUT should be a number
+    And save STDOUT as {POST_ID}
+
+    When I run `wp post update {POST_ID} --post_category={SECOND_TERM_ID}`
+    Then STDOUT should be:
+      """
+      Success: Updated post {POST_ID}.
+      """
+
+    When I run `wp post update {POST_ID} --post_category=test --porcelain`
+    Then STDOUT should be:
+      """
+      Warning: Unable to set test as post category
+      Success: Updated post {POST_ID}.
+      """
+
   Scenario: Creating/getting/editing posts
     Given a content.html file:
       """
--- a/features/post.feature
+++ b/features/post.feature
@@ -44,11 +44,26 @@ Feature: Manage WordPress posts
       Success: Updated post {POST_ID}.
       """
 
-    When I run `wp post update {POST_ID} --post_category=test --porcelain`
-    Then STDOUT should be:
+    When I try `wp post update {POST_ID} --post_category=test`
+    Then STDERR should be:
       """
       Warning: Unable to set test as post category
+      """
+    And STDOUT should be:
+      """
       Success: Updated post {POST_ID}.
+      """
+
+    When I try `wp post create --post_title="Test category two" --post_category={SECOND_TERM_ID},Test --porcelain`
+    Then STDOUT should be a number
+    And save STDOUT as {POST_ID_TWO}
+    Then STDERR should be:
+      """
+      Warning: Unable to set Test as post category
+      """
+    And STDOUT should be:
+      """
+      {POST_ID_TWO}
       """
 
   Scenario: Creating/getting/editing posts
--- a/features/post.feature
+++ b/features/post.feature
@@ -29,6 +29,7 @@ Feature: Manage WordPress posts
     When I try the previous command again
     Then the return code should be 1
 
+  Scenario: Setting post categories
     When I run `wp term create category "First Category" --porcelain`
     And save STDOUT as {TERM_ID}
     And I run `wp term create category "Second Category" --porcelain`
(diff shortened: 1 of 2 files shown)


schlessera:
@thrijith Can you provide the corresponding tests with the above change?

thrijith:
ok @schlessera

features/post.feature
```diff
@@ -29,6 +29,28 @@ Feature: Manage WordPress posts
...
+    Then STDOUT should be a number
+    And save STDOUT as {POST_ID}
+
+    When I run `wp post update {POST_ID} --post_category={SECOND_TERM_ID}`
+    Then STDOUT should be:
+      """
+      Success: Updated post {POST_ID}.
+      """
+
+    When I run `wp post update {POST_ID} --post_category=test --porcelain`
```
gitlost: If a command produces a warning or an error, it appears on `STDERR`, and need to use `try` instead of `run`:
```
    When I try `wp post update {POST_ID} --post_category=test`
    Then STDERR should be:
      """
      Warning: Unable to set test as post category
      """
    And STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """
```

gitlost: (This is a real gotcha with the behat tests, and we should throw a reminder to use `try` when it happens as it's so easy to forget to do it...)

gitlost:
@schlessera yes, agree that would be the preferable behavior.

gitlost:
From @thrijith slack:

> not sure i got it completely `https://github.com/wp-cli/entity-command/pull/129#issuecomment-357878268`
if a category term query for that string returns 0 IDs, throw an error that the category is unknown [done ?]
if a category term query for that string only returns 1 ID, use that ID instead [done ?]
if a category term query for that string returns more than 1 ID, throw an error that the category is ambiguous and that an ID is needed instead. [in what conditions will we get more than 1 ID ]

Yes you're right, `category_exists()` won't return more than 1 ID, only an int (if the category exists) or a falsey value (`0` or `null`), so the code in the loop would be eg:
```
$post_category = is_numeric( $post_category ) && (int) $post_category ? (int) $post_category : $post_category;
$category_id = category_exists( $post_category );
if ( ! $category_id ) {
	WP_CLI::error( "No such post category '$post_category'." );
}
$category_ids[] = $category_id;
```

gitlost:
Thinking about it this is probably better (as keep original passed-in `$post_category` to report in error):
```
if ( is_numeric( $post_category ) && (int) $post_category ) {
	$category_id = category_exists( (int) $post_category );
} else {
	$category_id = category_exists( $post_category );
}
```
etc.

features/post.feature
```diff
@@ -29,6 +29,33 @@ Feature: Manage WordPress posts
     When I try the previous command again
     Then the return code should be 1
 
+    When I run `wp term create category "First Category" --porcelain`
```
gitlost: I think it'd be better to move the category stuff to its own scenario. Here's an iteration that also checks some BC behaviour (for details, see `get_category_ids()` return comment below): 
```
  Scenario: Setting post categories
    When I run `wp term create category "First Category" --porcelain`
    And save STDOUT as {TERM_ID}
    And I run `wp term create category "Second Category" --porcelain`
    And save STDOUT as {SECOND_TERM_ID}

    When I run `wp post create --post_title="Test category" --post_category="First Category" --porcelain`
    Then STDOUT should be a number
    And save STDOUT as {POST_ID}

    When I run `wp post term list {POST_ID} category --field=term_id`
    Then STDOUT should be:
      """
      {TERM_ID}
      """

    When I run `wp post update {POST_ID} --post_category={SECOND_TERM_ID}`
    Then STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """

    When I run `wp post term list {POST_ID} category --field=term_id`
    Then STDOUT should be:
      """
      {SECOND_TERM_ID}
      """

    When I run `wp post update {POST_ID} --post_category='Uncategorized,{TERM_ID},Second Category'`
    Then STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """

    When I run `wp post term list {POST_ID} category --field=term_id`
    And save STDOUT as {MULTI_CATEGORIES_STDOUT}
    Then STDOUT should contain:
      """
      {TERM_ID}
      """
    And STDOUT should contain:
      """
      {SECOND_TERM_ID}
      """
    And STDOUT should contain:
      """
      1
      """

    # Blank categories with non-blank ignored.
    When I run `wp post update {POST_ID} --post_category='Uncategorized, ,{TERM_ID},Second Category,'`
    Then STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """

    When I run `wp post term list {POST_ID} category --field=term_id`
    Then STDOUT should be:
      """
      {MULTI_CATEGORIES_STDOUT}
      """

    # Zero category same as default Uncategorized (1) category.
    When I try `wp post update {POST_ID} --post_category=0`
    Then STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """

    When I run `wp post term list {POST_ID} category --field=term_id`
    Then STDOUT should be:
      """
      1
      """

    # Blank category/categories same as default Uncategorized (1) category.
    When I try `wp post update {POST_ID} --post_category=,`
    Then STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """

    When I run `wp post term list {POST_ID} category --field=term_id`
    Then STDOUT should be:
      """
      1
      """

    # Null category same as no categories.
    When I try `wp post update {POST_ID} --post_category=' '`
    Then STDOUT should be:
      """
      Success: Updated post {POST_ID}.
      """

    When I run `wp post term list {POST_ID} category --field=term_id`
    Then STDOUT should be empty

    # Non-existent category.
    When I try `wp post update {POST_ID} --post_category=test`
    Then STDERR should be:
      """
      Error: No such post category 'test'.
      """

    When I try `wp post create --post_title="Non-existent Category" --post_category={SECOND_TERM_ID},Test --porcelain`
    Then STDERR should be:
      """
      Error: No such post category 'Test'.
      """

    # Error on first non-existent category found.
    When I try `wp post create --post_title="More than one non-existent Category" --post_category={SECOND_TERM_ID},Test,Bad --porcelain`
    Then STDERR should be:
      """
      Error: No such post category 'Test'.
      """
```

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
...
 	}
+
+	/**
+	 * Get category id if given name/slug
+	 *
+	 * @param string $arg Supplied argument.
+	 * @return array
+	 */
+	private function get_category_ids( $arg ) {
+		$categoires = explode( ',', $arg );
```
gitlost: Typo `$categoires` -> `$categories`

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
...
+
+		foreach ( $categoires as $post_category ) {
```
gitlost: Need to init `$category_ids = array();` before loop.

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
...
+			if ( trim( $post_category ) !== '' ) {
```
gitlost: Think leave the comparison out, to just have `if ( trim( $post_category ) ) {`

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
...
+				if ( is_numeric( $post_category ) && (int) $post_category ) {
+					$category_id = category_exists( (int) $post_category );
+				} else {
+					$category_id = category_exists( $post_category );
+				}
+				if ( ! $category_id ) {
+					$category_ids[] = $post_category;
```
gitlost: This `$category_ids[] = $post_category;` can be removed as about to error and not return.

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
...
+					WP_CLI::error( "No such post category '$post_category'." );
+				} else {
```
gitlost: As `WP_CLI::error()` doesn't return, no need to else here.

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
...
+					$category_ids[] = $category_id;
+				}
+			}
+		}
+		return $category_ids;
```
gitlost: For BC with previous behaviour, we need to do:
```
// If no category ids found, return exploded array for compat with previous WP-CLI versions.
return $category_ids ? $category_ids : $categories;
```
here.

The BC behaviour is weird and non-intuitive, in that `--post_category=' '` will set no categories, while `--post_category=0` will set the default category (see tests above), but probably we should keep this behaviour to be safe...??

src/Post_Command.php
```diff
@@ -748,4 +748,31 @@ private function read_from_file_or_stdin( $arg ) {
 		}
 		return file_get_contents( $readfile );
 	}
+
+	/**
+	 * Get category id if given name/slug
```
gitlost: Maybe `Resolves post_category arg into an array of category ids.` or something better?

src/Post_Command.php
thrijith: @gitlost  tests look really good.
so if valid categories are passed we will send the resolved ids 
if a non-existent one is passed then throw an error

gitlost: Thanks @thrijith !

>so if valid categories are passed we will send the resolved ids
if a non-existent one is passed then throw an error ?

I think so, giving an error is a BC break but makes more sense than just warning...

gitlost:
Thanks @thrijith.

@schlessera could you give this a quick check before it's merged?

gitlost:
Merged this anyway as @schlessera has got enough on his plate. Thanks @thrijith !

schlessera:
I quickly looked through it and LGTM!

PR #133, opened 2018-01-19
Implement `--meta_input` with `post create` & `post update`

schlessera:
This needs https://github.com/wp-cli/wp-cli/pull/4623 to land first.

See https://github.com/wp-cli/wp-cli/pull/4616

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -162,6 +162,9 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 			$assoc_args['post_category'] = explode( ',', $assoc_args['post_category'] );
 		}
 
+		$array_arguments = array( 'tax_input', 'meta_input' );
+		$assoc_args      = \WP_CLI\Utils\parse_shell_arrays( $assoc_args, $array_arguments );
+
 		$assoc_args = wp_slash( $assoc_args );
 		parent::_create( $args, $assoc_args, function ( $params ) {
 			return wp_insert_post( $params, true );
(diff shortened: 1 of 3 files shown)


src/Post_Command.php
```diff
@@ -143,6 +143,10 @@ public function __construct() {
 	 *     # Create post with content from given file
 	 *     $ wp post create ./post-content.txt --post_category=201,345 --post_title='Post from file'
 	 *     Success: Created post 1922.
+	 *
+	 *     # Create a post with multiple meta values.
+	 *     $ wp post create --post_title='A post' --post_content='Just a small post.' --meta_input='{"key1":"value1","key2":"value2"}
```
Mr-Alexander: The `'` is not balanced here.

`--meta_input='{"key1":"value1","key2":"value2"}'`

src/Post_Command.php
```diff
@@ -271,6 +275,10 @@ public function create( $args, $assoc_args ) {
 	 *
 	 *     $ wp post update 123 --post_name=something --post_status=draft
 	 *     Success: Updated post 123.
+	 *
+	 *     # Update a post with multiple meta values.
+	 *     $ wp post update 123 --meta_input='{"key1":"value1","key2":"value2"}
```
Mr-Alexander: The ' is not balanced here.

it should be :
`--meta_input='{"key1":"value1","key2":"value2"}'`

swissspidy: Good catch, thanks!

src/Post_Command.php
swissspidy: This has been fixed a while ago in #147.

PR #144, opened 2018-02-12
Add subcommand for checking a user's password

### LINKED ISSUE #110: Add subcommand for checking a user's password
Hello,

I have a question about possible new subcommand for 'user' command before i will try to upstream it.

We (in Plesk) are making an WordPress Toolkit and we have "login" button functionality.  Before log in to WordPress we check admin credentials (stored in own db). User will be redirected to admin's credentials page if stored credentials doesn't work. We are using a self-written command for this purpose but this command could be easily moved to 'user' mainstream command:
`user check-password <login> <password>`

Is there any chance that such request will be approved?

thrijith:
fix #110 
add sub-command to validate user credentials.

### DIFF
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -1197,4 +1197,41 @@ class User_Command extends \WP_CLI\CommandWithDBObject {
 		Utils\report_batch_operation_results( 'user', $verb, count( $user_ids ), $successes, $errors );
 	}
 
+	/**
+	 * Checks if given user credentials are valid or not.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <user>
+	 * : The user login or user email of the user to check credentials.
+	 *
+	 * <user_pass>
+	 * : A string that contains the plain text password for the user.
+	 *
+	 * ## EXAMPLES
+	 *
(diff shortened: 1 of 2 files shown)


src/User_Command.php
```diff
@@ -1197,4 +1197,41 @@ private function update_msuser_status( $user_ids, $pref, $value ) {
 		Utils\report_batch_operation_results( 'user', $verb, count( $user_ids ), $successes, $errors );
 	}
 
+	/**
+	 * Checks if given user credentials are valid or not.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <user>
+	 * : The user login or user email of the user to check credentials.
```
gitlost: Should mention ID as well so `The user login, user email or user ID of the user to check credentials for.`

src/User_Command.php
```diff
@@ -1197,4 +1197,41 @@ private function update_msuser_status( $user_ids, $pref, $value ) {
 		Utils\report_batch_operation_results( 'user', $verb, count( $user_ids ), $successes, $errors );
 	}
 
+	/**
+	 * Checks if given user credentials are valid or not.
```
gitlost: Actually think this would read better as `Checks if a user's credentials are valid or not.`

PR #145, opened 2018-02-14
Fix bug when checking for default option value.

felixarntz:
Since WordPress 4.7.0, WordPress has had an advanced `register_setting()` function that among other improvements allows to specify a default value. This value will then automatically returned by unsuccessful `get_option()` calls when no actual `$default` parameter was provided (checked via `func_num_args()`).

Because of this change, the feedback can easily go wrong at the moment, since the command only checks for `false`, while the default value could also be something entirely different. By actually passing `false` as the default, it is ensured that the default value detection triggers correctly.

### DIFF
--- a/src/Option_Command.php
+++ b/src/Option_Command.php
@@ -73,7 +73,7 @@ class Option_Command extends WP_CLI_Command {
 	public function get( $args, $assoc_args ) {
 		list( $key ) = $args;
 
-		$value = get_option( $key );
+		$value = get_option( $key, false );
 
 		if ( false === $value ) {
 			WP_CLI::error( "Could not get '$key' option. Does it exist?" );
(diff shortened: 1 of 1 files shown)


schlessera:
Would that change relative to default WordPress behavior, though?

The question is, if we ask for an option, like for example the support email, and we get back a default value instead of `false`, like `support@example.com`... does this not mean we have a valid option value (even if it was not stored in the database)?

felixarntz:
That is a good question. It depends on whether we regard WP-CLI here in a way that it should check for an actual option value or whether it should also take the definition into account. I'd need to think further, but I think I do prefer including the registered default (in which case this could be closed).

danielbachhuber:
WP-CLI should follow WordPress' behavior as closely as possible, and only deviate when necessary.

danielbachhuber:
Closing to keep the existing behavior.

Issue #141, opened 2018-02-09
Request: Add `wp post duplicate`


jgraup:
It would be great to include a command that is almost identical to [`wp post create`](https://developer.wordpress.org/cli/commands/post/create/) except all the default values are pulled from an existing post (including meta data) and any additional commands will override those defaults.

```
wp post duplicate 123 --post_title="A Different Title"
```

gitlost:
Sounds good to me @jgraup !

schlessera:
If the command is almost identical, it might make more sense to implement this as an extension of `wp post create` instead.

Maybe something like this:
```
wp post create --from-post=123
```

sagarnasit:
I created a new command that takes id of existing post and generates a new post with metadata. Also added other options to override defaults. I don't know what kind of testing I should do and how. can anybody help?

eg. wp post **duplicate** 21

schlessera:
@sagarnasit I'd prefer to have this functionality be an extension of the existing `wp post create` command, as the functionality is almost identical.

aaronjorbin:
I think this can be closed due to #154

schlessera:
Fixed via #154

PR #153, opened 2018-03-06
add wp post duplicate subcommand

sagarnasit:
I created a separate subcommand for duplicating a post by providing post id. Related metadata will be duplicated too with the new post. a user will be outputted with a new post id. It accepts post id as an argument. A duplicate post data can be overridden by providing `[options]` with expecting values.

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -313,7 +313,135 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 		} );
 	}
 
-	/**
+      /**
+       * Create a duplicate post.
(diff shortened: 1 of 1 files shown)


schlessera:
@sagarnasit Thanks for submitting a pull request!

I'm very sorry, however, that I need to close this. As already stated in https://github.com/wp-cli/entity-command/issues/141 , I'd prefer to have the "duplication" functionality be an extension of "post create" instead of a separate command. These share almost their entire code, so having two separate commands just duplicates maintenance work and bug opportunities.

PR #154, opened 2018-03-14
Add `--from-post=<post_id>` flag to create duplicate posts

sagarnasit:
I here modified the create command to generate a duplicate post.

### DIFF
--- a/features/post-create-duplicate.feature
+++ b/features/post-create-duplicate.feature
@@ -1,0 +1,65 @@
+Feature: Create Duplicate WordPress post from existing posts.
+
+  Background:
+	Given a WP install
+
+  Scenario: Generate duplicate post.
+	When I run `wp term create category "Test Category" --porcelain`
+	Then save STDOUT as {TERM_ID}
+
+	When I run `wp post create --post_title='Test Duplicate Post' --post_category={TERM_ID} --porcelain`
+	And save STDOUT as {POST_ID}
+
+	When I run `wp post create --from-post={POST_ID} --porcelain`
+	Then STDOUT should be a number
+	And save STDOUT as {DUPLICATE_POST_ID}
+
+	When I run `wp post get {DUPLICATE_POST_ID} --field=title`
+	Then STDOUT should be:
+	  """
+	  Test Duplicate Post
+	  """
+
+	When I run `wp post term list {DUPLICATE_POST_ID} category --field=term_id`
+	Then STDOUT should be:
+      """
+      {TERM_ID}
+      """
+
+  Scenario: Generate duplicate post with post metadata.
+	When I run `wp post create --post_title='Test Post' --meta_input='{"key1":"value1","key2":"value2"}' --porcelain`
+	Then save STDOUT as {POST_ID}
+
+	When I run `wp post create --from-post={POST_ID} --porcelain`
+	Then save STDOUT as {DUPLICATE_POST_ID}
+
+	When I run `wp post meta list {DUPLICATE_POST_ID} --format=table`
+	Then STDOUT should be a table containing rows:
+	  | post_id             | meta_key | meta_value |
+	  | {DUPLICATE_POST_ID} | key1     | value1     |
+	  | {DUPLICATE_POST_ID} | key2     | value2     |
+
+  Scenario: Generate duplicate page.
+	When I run `wp post create --post_type="page" --post_title="Test Page" --post_content="Page Content" --porcelain`
+	Then save STDOUT as {POST_ID}
+
+	When I run `wp post create --from-post={POST_ID} --post_title="Duplicate Page" --porcelain`
+	Then save STDOUT as {DUPLICATE_POST_ID}
+
+	When I run `wp post list --post_type='page' --fields="title, content, type"`
+	Then STDOUT should be a table containing rows:
+	  | post_title     | post_content | post_type |
+	  | Test Page      | Page Content | page      |
+	  | Duplicate Page | Page Content | page      |
+
+  Scenario: Change type of duplicate post.
+	When I run `wp post create --post_title='Test Post' --porcelain`
+	Then save STDOUT as {POST_ID}
+
+	When I run `wp post create --from-post={POST_ID} --post_type=page --porcelain`
+	Then save STDOUT as {DUPLICATE_POST_ID}
+
+	When I run `wp post get {DUPLICATE_POST_ID} --fields=type`
+	Then STDOUT should be a table containing rows:
+	  | Field     | Value |
+	  | post_type | page  |
--- a/features/post-create-duplicate.feature
+++ b/features/post-create-duplicate.feature
@@ -7,8 +7,17 @@ Feature: Create Duplicate WordPress post from existing posts.
 	When I run `wp term create category "Test Category" --porcelain`
 	Then save STDOUT as {TERM_ID}
 
+	When I run `wp term create post_tag "Test Tag" --porcelain`
+	Then save STDOUT as {TAG_ID}
+
 	When I run `wp post create --post_title='Test Duplicate Post' --post_category={TERM_ID} --porcelain`
 	And save STDOUT as {POST_ID}
+
+	When I run `wp post term add {POST_ID} post_tag {TAG_ID} --by=id`
+	Then STDOUT should contain:
+      """*
+      Success: Added term.
+      """
 
 	When I run `wp post create --from-post={POST_ID} --porcelain`
 	Then STDOUT should be a number
@@ -25,6 +34,13 @@ Feature: Create Duplicate WordPress post from existing posts.
       """
       {TERM_ID}
       """
+
+	When I run `wp post term list {DUPLICATE_POST_ID} post_tag --field=term_id`
+	Then STDOUT should be:
+      """
+      {TAG_ID}
+      """
+
   @require-wp-4.4
   Scenario: Generate duplicate post with post metadata.
 	When I run `wp post create --post_title='Test Post' --meta_input='{"key1":"value1","key2":"value2"}' --porcelain`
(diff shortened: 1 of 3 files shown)


src/Post_Command.php
```diff
@@ -77,7 +77,10 @@ public function __construct() {
 	 *
 	 * [--post_name=<post_name>]
 	 * : The post name. Default is the sanitized post title when creating a new post.
-	 *
+       *
+       * [--from-post=<post_id>]
+       * : The post id of a post to be duplicated.
+       *
```
danielbachhuber: Looks like this is spaces instead of tabs. Can you correct to tabs?

src/Post_Command.php
```diff
@@ -147,7 +150,11 @@ public function __construct() {
 	 *     # Create a post with multiple meta values.
 	 *     $ wp post create --post_title='A post' --post_content='Just a small post.' --meta_input='{"key1":"value1","key2":"value2"}
 	 *     Success: Created post 1923.
-	 */
+       *
+       *     # Create a duplicate post with same post data.
+       *     $ wp post create --from-post=123 --post_title='Different Title'
+       *     Success: Created post 2350.
+       */
```
danielbachhuber: Ditto spaces -> tabs.

src/Post_Command.php
```diff
@@ -173,6 +180,21 @@ public function create( $args, $assoc_args ) {
 		$array_arguments = array( 'meta_input' );
 		$assoc_args      = \WP_CLI\Utils\parse_shell_arrays( $assoc_args, $array_arguments );
 
+            if( isset( $assoc_args['from-post'] ) ) {
+                  $post     = $this->fetcher->get_check( $assoc_args['from-post'] );
+                  $post_arr = get_object_vars( $post );
+                  $post_id  = $post_arr['ID'];
+                  unset( $post_arr['post_date'] );
+                  unset( $post_arr['post_date_gmt'] );
+                  unset( $post_arr['guid'] );
```
danielbachhuber: WHy did you decide to skip these values?

src/Post_Command.php
```diff
@@ -173,6 +180,21 @@ public function create( $args, $assoc_args ) {
 		$array_arguments = array( 'meta_input' );
 		$assoc_args      = \WP_CLI\Utils\parse_shell_arrays( $assoc_args, $array_arguments );
 
+            if( isset( $assoc_args['from-post'] ) ) {
+                  $post     = $this->fetcher->get_check( $assoc_args['from-post'] );
+                  $post_arr = get_object_vars( $post );
+                  $post_id  = $post_arr['ID'];
```
danielbachhuber: `$post_arr['ID']` should be unset, correct?

src/Post_Command.php
```diff
@@ -797,4 +819,22 @@ private function get_category_ids( $arg ) {
...
 		return $category_ids ? $category_ids : $categories;
 	}
+
+      /**
+       * Get post metadata.
+       *
+       * @param $post_id id of the post.
+       * @return array
+       */
+      private function get_metadata( $post_id = null ) {
```
danielbachhuber: Instead of making `$post_id` optional, we should make it required:

```
get_metadata( $post_id = null )
```

src/Post_Command.php
```diff
@@ -173,6 +180,21 @@ public function create( $args, $assoc_args ) {
...
+                  $post     = $this->fetcher->get_check( $assoc_args['from-post'] );
+                  $post_arr = get_object_vars( $post );
+                  $post_id  = $post_arr['ID'];
+                  unset( $post_arr['post_date'] );
+                  unset( $post_arr['post_date_gmt'] );
+                  unset( $post_arr['guid'] );
+
+                  if ( empty( $assoc_args['meta_input'] ) ) {
+                        $assoc_args['meta_input'] = $this->get_metadata( $post_id );
+                  }
```
danielbachhuber: Do we need to persist taxonomy data too?

src/Post_Command.php
sagarnasit: yes, I will correct it.

src/Post_Command.php
sagarnasit: yes, you are right. Actually, if we don't unset id it will still generate new id. But, for a safer side, I will correct.

src/Post_Command.php
sagarnasit: This values will be duplicated too. I skip post_date and post_date_gmt so that values can be reset according to the current date at the time of duplicating. Reason to unset guid is that duplicated post guid will same as the original.

If you have any suggestions please let me know.

src/Post_Command.php
sagarnasit: This part is for overwriting existing metadata if a user wants. Categories and tags can be duplicated too. what are your thoughts about it?

danielbachhuber: Yes, I think categories and tags should be duplicated too.

src/Post_Command.php
sagarnasit: sure.

sagarnasit:
@danielbachhuber I made changes you asked and added taxonomy duplication as well. I don't have any idea about functional testing so I haven't done testing yet.

schlessera:
@sagarnasit I can help you go through setting up the testing.

You should first check whether you can even run the existing tests.

As the Functional Tests help states, you'll need to prepare your database server first:

> Before running the functional tests, you’ll need a MySQL (or MariaDB) user called wp_cli_test with the password password1 that has full privileges on the MySQL database wp_cli_test. Running the following as root in MySQL should do the trick:

	GRANT ALL PRIVILEGES ON wp_cli_test.* TO "wp_cli_test"@"localhost" IDENTIFIED BY "password1";

> Then, to run the entire test suite:

	./vendor/bin/behat --expand

> Or to test a single feature:

	./vendor/bin/behat features/core.feature

schlessera:
Looking into it: https://github.com/wp-cli/entity-command/issues/161

schlessera:
@sagarnasit I should have addressed the test failure on PHP 5.3 now. Can you merge latest master into your PR to make the tests pass again?

sagarnasit:
@danielbachhuber I added functional testing as you said.

features/post-create-duplicate.feature
```diff
@@ -0,0 +1,66 @@
...
+	  """
+	  Test Duplicate Post
+	  """
+
+	When I run `wp post term list {DUPLICATE_POST_ID} category --field=term_id`
+	Then STDOUT should be:
+      """
+      {TERM_ID}
+      """
+  @require-wp-4.4
```
schlessera: Blank line needed between scenarios (above the `@require-wp-4.4` tag).

src/Post_Command.php
```diff
@@ -171,7 +178,28 @@ public function create( $args, $assoc_args ) {
 		}
 
 		$array_arguments = array( 'meta_input' );
-		$assoc_args      = \WP_CLI\Utils\parse_shell_arrays( $assoc_args, $array_arguments );
+		$assoc_args = \WP_CLI\Utils\parse_shell_arrays($assoc_args, $array_arguments);
```
schlessera: This entire block (lines 181-202) seems to have been inadvertently reformatted to PSR2 with the last commit.

Please adapt the formatting to adhere to WPCS instead.

src/Post_Command.php
```diff
@@ -797,4 +825,60 @@ private function get_category_ids( $arg ) {
...
+	}
+
+	/**
+	 * Get Tags of a post.
+	 *
+	 * @param $post_id postid of the post.
+	 *
+	 * @return array
+	 */
+	private function get_tags( $post_id ) {
```
schlessera: The code path for tags is not covered by the tests.

schlessera:
@sagarnasit This looks good already. There's some whitespace that got messed up during the last commit that you need to redress, and I would like to also have a test that covers the "tags" code.

src/Post_Command.php
```diff
@@ -797,4 +825,60 @@ private function get_category_ids( $arg ) {
...
+			array_push( $category_arr, $cat->term_id );
+		}
+
+		return $category_arr;
+	}
+
+	/**
+	 * Get Tags of a post.
+	 *
+	 * @param $post_id postid of the post.
```
schlessera: Should be:
```
@param $post_id ID of the post.
```

src/Post_Command.php
```diff
@@ -797,4 +825,60 @@ private function get_category_ids( $arg ) {
...
+			}
+		}
+
+		return $items;
+	}
+
+	/**
+	 * Get Categories of a post.
+	 *
+	 * @param $post_id postid of the post.
```
schlessera: Should be:
```
@param $post_id ID of the post.
```

src/Post_Command.php
```diff
@@ -797,4 +825,60 @@ private function get_category_ids( $arg ) {
 		// If no category ids found, return exploded array for compat with previous WP-CLI versions.
 		return $category_ids ? $category_ids : $categories;
 	}
+
+	/**
+	 * Get post metadata.
+	 *
+	 * @param $post_id id of the post.
```
schlessera: Should be:
```
@param $post_id ID of the post.
```

features/post-create-duplicate.feature
```diff
@@ -0,0 +1,82 @@
...
+
+	When I run `wp term create post_tag "Test Tag" --porcelain`
+	Then save STDOUT as {TAG_ID}
+
+	When I run `wp post create --post_title='Test Duplicate Post' --post_category={TERM_ID} --porcelain`
+	And save STDOUT as {POST_ID}
+
+	When I run `wp post term add {POST_ID} post_tag {TAG_ID} --by=id`
+	Then STDOUT should contain:
+      """*
```
schlessera: There's a random asterisk added here.

sagarnasit:
@schlessera Made changes you asked for.

schlessera:
Thank you for the pull request, @sagarnasit !

PR #155, opened 2018-03-16
Optional --skip-email flag for 'wp user update'

JanVoracek:
Sometimes (e.g. in our use case) it might be handy to update the user's details without bothering him with an email.

I would write some tests but I'm not really sure how to test it. I was looking for tests of `wp core install --skip-email` and they are missing as well.

### DIFF
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -501,13 +501,19 @@ class User_Command extends \WP_CLI\CommandWithDBObject {
 			$user_ids[] = $user->ID;
 		}
 
-		if ( Utils\get_flag_value( $assoc_args, 'skip-email' ) ) {
+		$skip_email = Utils\get_flag_value( $assoc_args, 'skip-email' );
+		if ( $skip_email ) {
 			add_filter( 'send_email_change_email', '__return_false' );
 			add_filter( 'send_password_change_email', '__return_false' );
 		}
 
 		$assoc_args = wp_slash( $assoc_args );
 		parent::_update( $user_ids, $assoc_args, 'wp_update_user' );
+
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> I would write some tests but I'm not really sure how to test it.

Here's an example: https://github.com/wp-cli/entity-command/blob/4933b26dc5f64631a85997281a25bbb6e397e42f/features/user-reset-password.feature#L16

src/User_Command.php
```diff
@@ -498,6 +501,11 @@ public function update( $args, $assoc_args ) {
 			$user_ids[] = $user->ID;
 		}
 
+		if ( isset( $assoc_args['skip-email'] ) ) {
```
danielbachhuber: This should use `WP_CLI\Utils\get_flag_value()`

src/User_Command.php
```diff
@@ -498,6 +501,11 @@ public function update( $args, $assoc_args ) {
...
+			add_filter( 'send_email_change_email', '__return_false' );
+			add_filter( 'send_password_change_email', '__return_false' );
```
danielbachhuber: Can we remove these filters after the fact, to ensure they don't bleed through to other commands?

danielbachhuber:
@JanVoracek Still planning to come around and fix this up?

JanVoracek:
Yes, I will fix it. Sorry for waiting.

schlessera:
@JanVoracek Are you still up for working on this? Do you need any further help with something?

schlessera:
Thanks for your work on this, @JanVoracek !

PR #162, opened 2018-04-02
Raise memory limit to 256MB

### LINKED ISSUE #161: Travis runs out of memory on PHP 5.3 tests
The current behat tests cause an out-of-memory error on PHP 5.3: https://travis-ci.org/wp-cli/entity-command/jobs/360848530

We need to bump Travis PHP memory, 256 MB should be a reasonable target for now.

schlessera:
Fixes #161

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,9 @@ before_install:
     else
       echo "xdebug.ini does not exist"
     fi
+  - |
+    # Raise PHP memory limit to 256MB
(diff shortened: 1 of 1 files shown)


schlessera:
I should have read the Travis log more carefully. I thought it was running out of memory at about ~154MB. However, it runs out of memory at 1536MB instead. The memory limit is just fine, but Composer seems to use an inordinate amount of memory in PHP 5.3 environment.

Just raising the memory limit will not do in this case.

PR #167, opened 2018-04-09
Improve the formatting of the option autoload parameter documentation

johnbillion:
Documents the default and improves formatting.


danielbachhuber:
The options are quoted to ensure the YAML parser interprets them as strings, which the options API requires.

Good thing we have tests!

johnbillion:
Hurrah! Just that default value is missing, in that case, which is really what I wanted to fix.

PR #168, opened 2018-04-09
Improve the formatting of various format parameter documentation

johnbillion:
I'm pretty sure that this is correct but it does need double-checking for accuracy. Maybe even tests.


danielbachhuber:
Need to run `wp scaffold package-readme` to pass these changes through. Doing so now.

Issue #158, opened 2018-03-21
Add wp site meta subcommand to manage blogmeta


spacedmonkey:
Now that blog meta has landed in [core](https://core.trac.wordpress.org/ticket/37923), there should be a way to manage it in WP CLI.

schlessera:
Yes, I agree. Do you already have specific thoughts about how this should be modeled?

schlessera:
Just to start the discussion, here's what it could look like if we modeled it after `wp post meta`:

- `wp site meta add <id> <key> [<value>] [--format=<format>]`
- `wp site meta delete <id> [<key>] [<value>] [--all]`
- `wp site meta get <id> <key> [--format=<format>]`
- `wp site meta list <id> [--keys=<keys>] [--fields=<fields>] [--format=<format>] [--orderby=<fields>] [--order=<order>]`
- `wp site meta patch <action> <id> <key> <key-path>... [<value>] [--format=<format>]`
- `wp site meta pluck <id> <key> <key-path>... [--format=<format>]`
- `wp site meta update <id> <key> [<value>] [--format=<format>]`

Does this all make sense based on the current `wp_blogmeta` implementation in trunk, or are there commands that are invalid?

spacedmonkey:
In #159 I have created a first implementation of this sub command. It is a pretty standard meta table, I have just implemented the existing meta class and extended. 

As for merging testing, blogmeta is in trunk, but this PR might be better to merge at 5.0.0 is out, as it easier to test. As is, we can't be use that table has been installed, which is the reason, I have added a check to `is_site_meta_supported` a new core function to check it is there.

schlessera:
The tests currently break because the version requirement of 5.0 is not met.

Wouldn't it be enough to check for the existence of the `is_site_meta_supported` function?

spacedmonkey:
In the test, I have a check to see if it WP 5.0.0  - https://github.com/wp-cli/entity-command/pull/159/files#diff-730a53ee55d6e1e534aa6972bb0802aeR3. This is how term meta [works](https://github.com/wp-cli/entity-command/blob/master/features/term-meta.feature#L3). 

I did those to that these tests would be skipped on earlier version and tests wouldn't fail. Not sure how else it should work, unless I am missing something.

spacedmonkey:
Version number in trunk is 5.0 - https://github.com/WordPress/WordPress/blob/master/wp-includes/version.php#L7

schlessera:
[...]

PR #159, opened 2018-03-22
Add wp site meta sub-command

spacedmonkey:
Add the wp site meta sub-command, with tests. 

This one is going to be hard to test and might be merge after 5.0.0 is released. 

fixes #158

PR #174, opened 2018-04-16
Abstract meta CRUD into methods

### LINKED ISSUE #172: Making changes to meta objects leaves behind stale object caches
The commands that manipulate meta data leave behind stale object caches.

I noticed this here: https://github.com/wp-cli/entity-command/issues/158#issuecomment-375959721

Basically, the `add_metadata()` call does not include the additional cache-busting call that `add_site_meta()` includes. The same problem applies for all `add`, `update` and `delete` calls on meta objects.

Most pragmatic solution:
Provide a default mapping of CRUD function in the base object that can be optionally  overridden in the specific meta commands.

spacedmonkey:
In this PR, I abstract off the meta crud into there own method. This means that can overridden from the parent class. 

fixes #172 
related: #158

### DIFF
--- a/src/Comment_Meta_Command.php
+++ b/src/Comment_Meta_Command.php
@@ -24,6 +24,23 @@
 class Comment_Meta_Command extends \WP_CLI\CommandWithMeta {
 	protected $meta_type = 'comment';
 
+
+	protected function add_metadata( $object_id, $meta_key, $meta_value, $unique = false ) {
+		return add_comment_meta( $object_id, $meta_key, $meta_value, $unique );
+	}
+
+	protected function update_metadata( $object_id, $meta_key, $meta_value, $prev_value = '' ) {
+		return update_comment_meta( $object_id, $meta_key, $meta_value, $prev_value );
+	}
+
+	protected function get_metadata( $object_id, $meta_key = '', $single = false ) {
+		return get_comment_meta( $object_id, $meta_key, $single );
+	}
+
+	protected function delete_metadata( $meta_type, $object_id, $meta_key, $meta_value = '', $delete_all = false ) {
+		return delete_comment_meta( $object_id, $meta_key, $meta_value, $delete_all );
+	}
+
 	/**
 	 * Check that the comment ID exists
 	 *
@@ -35,4 +52,3 @@ class Comment_Meta_Command extends \WP_CLI\CommandWithMeta {
 		return $comment->comment_ID;
 	}
 }
-
--- a/src/Term_Meta_Command.php
+++ b/src/Term_Meta_Command.php
@@ -37,4 +37,20 @@ class Term_Meta_Command extends \WP_CLI\CommandWithMeta {
 		return $term->term_id;
 	}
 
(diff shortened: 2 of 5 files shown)


src/WP_CLI/CommandWithMeta.php
```diff
@@ -1,4 +1,4 @@
-<?php
+$this->update_metadata( <?php
```
schlessera: There's nonsense up here... ;)

src/Comment_Meta_Command.php
```diff
@@ -24,6 +24,23 @@
 class Comment_Meta_Command extends \WP_CLI\CommandWithMeta {
 	protected $meta_type = 'comment';
 
+
```
schlessera: This should only have 1 blank line here.

src/Term_Meta_Command.php
```diff
@@ -37,4 +37,20 @@ protected function check_object_id( $object_id ) {
...
+	}
+
+	protected function get_metadata( $object_id, $meta_key = '', $single = false ) {
+		return get_term_meta( $object_id, $meta_key, $single );
+	}
+
+	protected function delete_metadata( $object_id, $meta_key, $meta_value = '', $delete_all = false ) {
+		return delete_term_meta( $object_id, $meta_key, $meta_value, $delete_all );
+	}
+
```
schlessera: Redundant blank line here.

schlessera:
Good approach, only a few style fixes and the removal of a random paste needed.

spacedmonkey:
I have added some basic comments to this. @schlessera let me know if you want me to change them, I am not great for writing comments. Otherwise, I am happy with the PR as it is. Don't think there needs more work on it. 

This PR needs to be merged before, https://github.com/wp-cli/entity-command/pull/159 as that PR will need to changed, to implement changes in this PR. 

The tests seem to be failing on trunk, but I think that is unreleated to my change...

schlessera:
Ah, it seems to break on trunk now because a default `Privacy Policy` page was added as part of the GDPR effort.

I'll fix the tests for that.

spacedmonkey:
My bad. That is a silly make. Sorry, added .idea to gitignore file.

schlessera:
@spacedmonkey I actually meant for you to add `.idea` to your own, global .gitignore file, not the project's one. The project's .gitignore should not need to care about the .idea folder, that is client-specific tooling.

See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

spacedmonkey:
@schlessera Tests are passing and feedback done.

schlessera:
@spacedmonkey I pushed a change to add full docblocks to the methods.

While doing so, I noticed that the functions `delete_comment_meta()`, `delete_post_meta()`, `delete_term_meta()` and `delete_user_meta()` only accept 3 arguments, not the 4 that are currently being passed.

Currently, the `WP_CLI\CommandWithMeta::delete_metadata()` method has a fourth argument `$delete_all = false` which is accepted by the general `delete_metadata()` function, but not by the specialized versions.

Reading through the docs, it seems to me that this argument does not make any sense in our current context, and would be more useful for a separate command.

How do you suggest we proceed?

PR #177, opened 2018-04-17
Fix tests after Core's introduction of a default "Privacy Policy" page

### LINKED ISSUE #176: Tests fail on added `Privacy Policy` page
The tests currently fail on `trunk` because a `Privacy Policy` page was added as part of the GDPR effort [[Changeset](https://core.trac.wordpress.org/changeset/42981)].

schlessera:
Fixes #176

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -311,6 +311,7 @@ Feature: Manage WordPress posts
     Then STDOUT should be:
       """
       Sample Page
+      Privacy Policy
       """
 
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@schlessera Looks like this is still failing? Do the tests need to be dynamic based on WP version?

schlessera:
@danielbachhuber Yes, it is. I quickly realized that I only fixed it for `trunk`, but we're not dealing with a normal web project targeting a single platform here. So I modified the fix, but lost the internet connection when the plane took off before being able to push.

It should hopefully work on non-`trunk` as well now. :)

Issue #180, opened 2018-04-28
Add an `exists` command


johnbillion:
I recently had a situation where a post's parent post didn't exist due to data corruption. While updating my script I found the need for a `wp post exists <id>` command, but there isn't one.

There is a `wp comment exists` command. The same for core entities would be handy.

schlessera:
Fixed via #182

PR #182, opened 2018-05-02
Add `post exists` command

BhargavBhandari90:
For https://github.com/wp-cli/entity-command/issues/180

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -897,7 +897,7 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 		return $tag_arr;
 	}
 
-	/*
+	/**
 	 * Verifies whether a post exists.
 	 *
 	 * Displays a success message if the post does exist.
@@ -919,7 +919,6 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 	 *     $ wp post exists 10000
 	 *     $ echo $?
 	 *     1
-	 *
 	 */
 	public function exists( $args ) {
 		if ( $this->fetcher->get( $args[0] ) ) {
--- a/README.md
+++ b/README.md
@@ -2177,12 +2177,42 @@ wp post edit <id>
 
 
 
+### wp post exists
+
+Verifies whether a post exists.
+
+~~~
+wp post exists <id>
+~~~
+
+Displays a success message if the post does exist.
+
+**OPTIONS**
+
+	<id>
+		The ID of the post to check.
+
+**EXAMPLES**
+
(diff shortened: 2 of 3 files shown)


schlessera:
I see that you modeled this after `wp comment exists`. That one is also not easily usable through scripting. I think it should be changed as well.

For BC reasons, you could print a `WP_CLI::success()` message for success (which produces an exit code of `0`) or `WP_CLI::halt()` on failure with an exit code of `1`.

schlessera:
@BhargavBhandari90 Are you still up for working on this? Do you need help with something?

BhargavBhandari90:
Hi @schlessera,

Please check my latest commit. Is this what you wanted? If yes then I will modify comment command and feature tests.

Sorry for the delay.

src/Post_Command.php
```diff
@@ -896,4 +896,38 @@ private function get_tags( $post_id ) {
...
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     # The post is exists.
+	 *     $ wp post exists 1
+	 *     Success: Post with ID 1337 exists.
+	 *     $ echo $?
+	 *     0
+	 *
+	 *     # The post is not exists.
```
swissspidy: Correct would be:

> The post does not exist.

BhargavBhandari90: oops!!!! Thanks @swissspidy

features/post.feature
```diff
@@ -8,6 +8,12 @@ Feature: Manage WordPress posts
     Then STDOUT should be a number
     And save STDOUT as {POST_ID}
 
+    When I run `wp post exists {POST_ID}`
+    Then STDOUT should be:
+      """
+    Success: Post with ID {POST_ID} exists.
+      """
+
```
schlessera: The exit code should also be tested here.

Also, you should add a test for the negative case as well.

features/post.feature
```diff
@@ -8,6 +8,17 @@ Feature: Manage WordPress posts
...
     And save STDOUT as {POST_ID}
 
+    When I run `wp post exists {POST_ID}`
+    Then STDOUT should be:
+      """
+    Success: Post with ID {POST_ID} exists.
+      """
+    And the return code should be 0
+
+    When I run `wp post exists 1000`
```
schlessera: As you expect this call to return an error (non-null exit code), you need to use `When I try...`, not `When I run...`

BhargavBhandari90:
@schlessera Requested changes are done. Sorry for not being so late.

PR #184, opened 2018-05-08
Post Generate: Add support for post_date_gmt

### LINKED ISSUE #183: Post Generate: Add support for post_date_gmt
Running `wp post generate` with a `--post_date` specified leads to records with zeroed `post_date_gmt` values.

For example, when I run: `wp post generate --count=1 --post_date=2018-05-02-00:00:00`

I end up w/ the following:
```
> wp post list --fields=post_date,post_date_gmt
+---------------------+---------------------+
| post_date           | post_date_gmt       |
+---------------------+---------------------+
| 2018-05-02 00:00:00 | 1970-01-01 00:00:00 |
+---------------------+---------------------+
```

The default value [should be](https://developer.wordpress.org/reference/functions/wp_insert_post/#parameters) set to that of `post_date`, but it would be nice to also be able to specify on the command line.

I [started](https://github.com/jblz/entity-command/commit/a0f3e2842196ac7f2e93d432d2963235d17c1c60) to open a PR, but noticed the [Contributing Guidelines](https://github.com/wp-cli/entity-command#creating-a-pull-request) suggested making an issue first, so here I am :)

jblz:
Sets the GMT / UTC date of the generated posts. 

If the `--post_date_gmt` flag is provided, use that.
If not, default to the value of post_date (which defaults to the current date if _it's_ not set).

Fixes #183

### DIFF
--- a/features/post-generate.feature
+++ b/features/post-generate.feature
@@ -112,3 +112,55 @@ Feature: Generate new WordPress posts
       """
       2018-07-04 00:00:00
       """
+
+  Scenario: Generating posts with post_date argument with hyphenated time
+    When I run `wp post generate --count=1 --post_date="2018-07-05-17:17:17"`
+    And I run `wp post list --field=post_date`
+    Then STDOUT should contain:
+      """
+      2018-07-05 17:17:17
+      """
+    And I run `wp post list --field=post_date_gmt`
+    Then STDOUT should contain:
+      """
+      2018-07-05 17:17:17
+      """
+
+  Scenario: Generating posts with post_date_gmt argument with hyphenated time
+    When I run `wp post generate --count=1 --post_date_gmt="2018-07-06-12:12:12"`
+    And I run `wp post list --field=post_date`
+    Then STDOUT should contain:
+      """
+      2018-07-06 12:12:12
+      """
+    And I run `wp post list --field=post_date_gmt`
+    Then STDOUT should contain:
+      """
+      2018-07-06 12:12:12
+      """
+
+  Scenario: Generating posts with different post_date & post_date_gmt argument without time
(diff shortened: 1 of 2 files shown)


jblz:
Needs Tests

schlessera:
@jblz Thanks for the pull request. Are you up for providing the tests as well?

jblz:
> Are you up for providing the tests as well?

@schlessera Happy to, but would love a pointer on where / how to add them.

I'm not seeing anything related [here](https://github.com/wp-cli/entity-command/tree/d44e0bfb09a63bafa5b1be0125ad619662b2f422/tests) for example and you can probably point me in the right direction more quickly than I can figure it out on my own :)

schlessera:
@jblz The tests are done as functional tests in Behat.

You can find the current tests for `wp post generate` here: https://github.com/wp-cli/entity-command/blob/d44e0bfb09a63bafa5b1be0125ad619662b2f422/features/post-generate.feature

You can find an overview of how to work on tests in WP-CLI here: https://make.wordpress.org/cli/handbook/pull-requests/#running-and-writing-tests

schlessera:
@jblz Are you still up for working on the tests? Do you need further help with that?

jblz:
@schlessera Thanks for the nudge.  I managed to get my test env set up & make this actually work :)

I think it's ready for review if the CI comes back green.

jblz:
> they already uncovered a bug that you needed to fix to make the tests pass, right?

Yep!  I needed to pull the `current_time` call out so we could tell if the caller was relying on defaults for `post_date` as well as `post_date_gmt` & do the correct thing in all cases.

> So it was definitely not a waste of time to write the tests.

It rarely is!

Requested changes are in place.

schlessera:
Thanks for the pull-request, @jblz !

PR #188, opened 2018-07-20
Delete alert in case of custom post type 

### LINKED ISSUE #128: wp post delete on custom post type
Hi! Recently I found annoying issue with `wp post delete` command. I tried to trash some woocommerce products using this command and found that products where deleted completely from database instead. After digging into source code I found that this command uses `wp_delete_post` with `$force_delete` parameter. This parameter is very specific and works as expected only for 'post' and 'page' post types and for custom post types `$force_delete` doesn't have effect and they are always deleted by force.

I believe this is quite deceiving and leads to unexpected behavior. Please fix.

Mte90:
Closes #128


Mte90:
Do we need a behat test for that?

Mte90:
From what I can see in `post.feature` there are post id but there is no way to detect the post type itself so I am not sure what to do for that part.

Mte90:
The tests are failing on the case of `force` and I don't understand why it's happening.
When I tried the command on a site was working, if someone can help on that will be appreciated :-)

Mte90:
thanks for the help!

schlessera:
I had trouble figuring out what the actual problem was... until I noticed that the failure came from a different test and it was just because the tests were running out of order.

Issue #150, opened 2018-02-23
Expose the available query variables for comment and post queries


johnbillion:
`wp user list --help` displays a helpful link to https://developer.wordpress.org/reference/classes/wp_user_query/prepare_query/

The same isn't true for comment and post queries. A link to the relevant docs on developer.wordpress.org would aid the discovery of the available query variables.

schlessera:
Yes, agreed. The available fields should be shown a bit below, though.

PR #213, opened 2018-10-04
Improve documentation

austinginder:
Improve docs for #150


schlessera:
These should normally be added to the `README.md` file as well, but the current scaffolding code does not seem to pick them up. I created a fresh issue for that here: https://github.com/wp-cli/scaffold-package-command/issues/181

In the mean time, let's just merge this and deal with updating the `README.md` file later.

Issue #190, opened 2018-08-02
Presentation of serialized and JSON-encoded meta values are ambiguous, behaviour differs from options


johnbillion:
I've just been tripped up by the way that WP-CLI represents serialized data stored in meta data as JSON encoded, and the fact that the behaviour when displaying meta data differs from that when displaying option values.

Scenario: Two post meta fields, one storing its data as a JSON-encoded string and one storing its data using the WordPress-native serialisation:

```php
$value = [ 1, 2, 3 ];
add_post_meta( $id, 'json', wp_json_encode( $value ) );
add_post_meta( $id, 'raw', $value );
```

When listing the post meta for this post, WP-CLI incorrectly represents both values as JSON-encoded:

```
wp post meta list <id>
+----------+------------+
| meta_key | meta_value |
+----------+------------+
| json     | [1,2,3]    |
| raw      | [1,2,3]    |
+----------+------------+
```

When doing the same with options instead of post meta, the serialised value is correctly represented when listed:

```
wp option list
+-------------+--------------------------------+
| option_name | option_value                   |
+-------------+--------------------------------+
| json        | [1,2,3]                        |
| raw         | a:3:{i:0;i:1;i:1;i:2;i:2;i:3;} |
+-------------+--------------------------------+
```

This means that when using the `list` command it's not possible to determine whether a meta value is actually stored serialised or JSON encoded. When doing the same for option values, it is.

IMO I think the representation of meta values should be brought inline with that of option values.

schlessera:
I assume this is the same for `comment meta list`, `term meta list`, `site meta list` and `user meta list`...

johnbillion:
Needs checking to confirm, but I assumed so.

BhargavBhandari90:
I am getting Fatal error while using site meta list command.

here is the error http://prntscr.com/l5fyby

johnbillion:
This was fixed in #215 and #254.

PR #215, opened 2018-10-12
Harmonize representation of serialized and JSON-encoded meta values

BhargavBhandari90:
For #190


schlessera:
Thanks for the pull-request, @BhargavBhandari90 !

I'm wondering right now whether we shouldn't consider backward compatibility first before making this change. I don't know how common it is for people to directly use the JSON representation they get from the meta here... I'll bring this up during office hours.

schlessera:
Added a new issue #245 which allows users to switch back to the previous behavior if needed.

schlessera:
Thanks for the PR, @BhargavBhandari90 !

Issue #216, opened 2018-10-19
User description is not saved in the database


infostreams:
If you create a new user, and want to provide a description for that user, this is how you are supposed to do it:

`wp user create slug email@example.com --description='A very fine user indeed'`

However, the description is never stored in the database. 

Turns out that the 'description' flag is simply not parsed. See https://github.com/wp-cli/entity-command/blob/master/src/User_Command.php and search for 'description'. You will get four matches, all of them in the docs, but none in the code. If you search for 'first_name', on the other hand, you will get some matches in the code as well.

BhargavBhandari90:
@infostreams Yes, you are right. It is not saving the description

PR #217, opened 2018-10-20
Fix - #216 user description was not getting saved

BhargavBhandari90:
Fix for #216


BhargavBhandari90:
@swissspidy Added test. I added to meta featured file. Hope that makes sense. Let me know if it needs to be improved.

PR #236, opened 2019-02-28
Reset option value to default for privacy policy on site empty

thrijith:
Fixes https://github.com/wp-cli/wp-cli/issues/5086

### DIFF
--- a/features/site-empty.feature
+++ b/features/site-empty.feature
@@ -29,6 +29,21 @@ Feature: Empty a WordPress site of its data
       Success: Created post_tag 2.
       """
 
+    When I run `wp post create --post_type=page --post_title='Sample Privacy Page' --post_content='Sample Privacy Terms' --porcelain`
+    Then save STDOUT as {PAGE_ID}
+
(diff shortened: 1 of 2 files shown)


thrijith:
Hi @danielbachhuber, I have added a test in an existing scenario. LMK if it's ok.

features/site-empty.feature
```diff
@@ -42,6 +42,12 @@ Feature: Empty a WordPress site of its data
     When I run `wp term list post_tag --format=ids`
     Then STDOUT should be empty
 
+    When I run `wp option get wp_page_for_privacy_policy`
+    Then STDOUT should be:
+      """
+      0
+      """
```
danielbachhuber: Can we also include a test that makes sure there's a non-zero value _before_ `wp site empty` is run?

thrijith: I am not sure how to do that, should I create a dummy page and assign it as privacy page?

danielbachhuber: > should I create a dummy page and assign it as privacy page?

Yes, this would be fine.

PR #241, opened 2019-03-13
Add count to post-type and taxonomy commands

atanas-angelov-dev:
Fixes wp-cli/wp-cli#4573.

I took the liberty of implementing the term count for the `taxonomy` command as well even though it was not mentioned in the original issue as it is in a very similar situation.

Feedback is very much welcome.

### DIFF
--- a/src/Post_Type_Command.php
+++ b/src/Post_Type_Command.php
@@ -35,6 +35,37 @@ class Post_Type_Command extends WP_CLI_Command {
 	);
 
 	/**
+	 * Gets the post counts for each supplied post type.
+	 *
+	 * @param array $post_types Post types to fetch counts for.
+	 * @return array Associative array of post counts keyed by post type.
+	 */
+	protected function get_counts( $post_types ) {
+		global $wpdb;
+
+		if ( count( $post_types ) <= 0 ) {
+			return [];
+		}
+
+		$query  = $wpdb->prepare(
+			"SELECT `post_type`, COUNT(*) AS `count`
+			FROM $wpdb->posts
+			WHERE `post_type` IN (" . implode( ',', array_fill( 0, count( $post_types ), '%s' ) ) . ")
+			GROUP BY `post_type`",
+			$post_types
+		);
+		$counts = $wpdb->get_results( $query );
+
+		// Make sure there's a count for every item.
+		$counts = array_merge(
+			array_fill_keys( $post_types, 0 ),
+			wp_list_pluck( $counts, 'count', 'post_type' )
+		);
+
+		return $counts;
+	}
+
+	/**
 	 * Lists registered post types.
 	 *
 	 * ## OPTIONS
@@ -71,7 +102,9 @@ class Post_Type_Command extends WP_CLI_Command {
 	 * * public
 	 * * capability_type
 	 *
-	 * There are no optionally available fields.
+	 * These fields are optionally available:
+	 *
+	 * * count
 	 *
 	 * ## EXAMPLES
 	 *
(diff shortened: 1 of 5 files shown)


atanas-angelov-dev:
The WP 3.7.11 test failed because the count for the `page` post type doesn't match up as only one page is created (Sample Page) vs 2 pages on latest (Sample Page and Privacy Policy).

I don't really have much experience with Behat so I'm not really sure what's the best approach here.

schlessera:
@atanas-angelov-dev First of all, thanks for the PR, This looks like a great start!

I'm still a bit wary of the performance impact of the count, so I suggest the following change:

- Don't add the `count` column to the output by default (also for BC reasons).
- Only execute the queries when the `count` column was actually included in the output using the `--fields` argument.

Regarding the failure on WP 3.7, I don't think we need to bother with testing the counts on there. Just add a `@require-wp-5.0` and we're good to go. Note though that you'll have to provide a separate test for the count functionality once it's not part of the default output anymore.

atanas-angelov-dev:
@schlessera feedback has been addressed 👍

I had to limit the taxonomy tests to 5.0 as well since category count is 0 instead of 1 on 3.7.11 (guess Uncategorized is not created automatically under test conditions on that version or something?)

src/Post_Type_Command.php
```diff
@@ -34,6 +34,37 @@ class Post_Type_Command extends WP_CLI_Command {
...
+	 * Gets the post counts for each supplied post type.
+	 *
+	 * @param array $post_types Post types to fetch counts for.
+	 * @return array Associative array of post counts keyed by post type.
+	 */
+	protected function get_counts( $post_types ) {
+		global $wpdb;
+
+		if ( count( $post_types ) <= 0 ) {
+			return array();
```
schlessera: We're at PHP 5.4+ now, so we should use short array syntax.
```suggestion
			return [];
```

src/Post_Type_Command.php
```diff
@@ -100,7 +131,18 @@ class Post_Type_Command extends WP_CLI_Command {
 	public function list_( $args, $assoc_args ) {
 		$formatter = $this->get_formatter( $assoc_args );
 
-		$types = get_post_types( $assoc_args, 'objects' );
+		$fields = $formatter->fields;
+		$types  = get_post_types( $assoc_args, 'objects' );
+		$counts = array();
```
schlessera:
```suggestion
		$counts = [];
```

src/Post_Type_Command.php
```diff
@@ -169,6 +211,15 @@ public function get( $args, $assoc_args ) {
 			$assoc_args['fields'] = $default_fields;
 		}
 
+		$formatter = $this->get_formatter( $assoc_args );
+		$fields    = $formatter->fields;
+		$count     = 0;
+
+		if ( in_array( 'count', $fields, true ) ) {
+			$count = $this->get_counts( array( $post_type->name ) );
```
schlessera:
```suggestion
			$count = $this->get_counts( [ $post_type->name ] );
```

src/Taxonomy_Command.php
```diff
@@ -44,6 +44,37 @@ public function __construct() {
...
+	 * Gets the term counts for each supplied taxonomy.
+	 *
+	 * @param array $taxonomies Taxonomies to fetch counts for.
+	 * @return array Associative array of term counts keyed by taxonomy.
+	 */
+	protected function get_counts( $taxonomies ) {
+		global $wpdb;
+
+		if ( count( $taxonomies ) <= 0 ) {
+			return array();
```
schlessera:
```suggestion
			return [];
```

src/Taxonomy_Command.php
```diff
@@ -112,10 +143,17 @@ public function list_( $args, $assoc_args ) {
 			$assoc_args['object_type'] = array( $assoc_args['object_type'] );
 		}
 
+		$fields     = $formatter->fields;
 		$taxonomies = get_taxonomies( $assoc_args, 'objects' );
+		$counts     = array();
```
schlessera:
```suggestion
		$counts     = [];
```

src/Taxonomy_Command.php
```diff
@@ -179,6 +217,15 @@ public function get( $args, $assoc_args ) {
 			$assoc_args['fields'] = $default_fields;
 		}
 
+		$formatter = $this->get_formatter( $assoc_args );
+		$fields    = $formatter->fields;
+		$count     = 0;
+
+		if ( in_array( 'count', $fields, true ) ) {
+			$count = $this->get_counts( array( $taxonomy->name ) );
```
schlessera:
```suggestion
			$count = $this->get_counts( [ $taxonomy->name ] );
```

atanas-angelov-dev:
@schlessera all done - I even ended up adjusting the default fields section for `Taxonomy_Command::list_()` as it was outdated.

PR #256, opened 2019-05-05
Add strict/no-strict mode for taxonomy list

### LINKED ISSUE #181: Use get_object_taxonomies() for taxonomy list by post type object
I was running the command from the example `wp taxonomy list --object_type=post --fields=name,public` and could not figure out why all of the taxonomies were not being displayed.

One of the taxonomies that was connected to the "post" object was also associated with other post types. Turns out this is one of the limitations([Core Trac](https://core.trac.wordpress.org/ticket/27918)) of `get_taxonomies()` which is used by [`wp taxonomy list`](https://github.com/wp-cli/entity-command/blob/master/src/Taxonomy_Command.php#L115).

A solution would be to use [`get_object_taxonomies()`](https://developer.wordpress.org/reference/functions/get_object_taxonomies/) instead.

I am not sure what the solution would be here. Perhaps add a different flag or add a new command.

BhargavBhandari90:
Added `strict` mode for taxonomy list command.
Default will be --no-strict

Fixes #181.

### DIFF
--- a/features/taxonomy.feature
+++ b/features/taxonomy.feature
@@ -46,3 +46,42 @@ Feature: Manage WordPress taxonomies
       | Field       | Value      |
       | name        | category   |
       | count       | 1          |
+
+  Scenario: Listing taxonomies with strict/no-strict mode
+    Given a WP installation
+    And a wp-content/mu-plugins/test-taxonomy-list.php file:
+      """
+      <?php
+      // Plugin Name: Test Taxonomy Strict/No-Strict Mode
+
+      add_action( 'init', function(){
+        $args = array(
+          'hierarchical'          => true,
+          'show_ui'               => true,
+          'show_admin_column'     => true,
+          'update_count_callback' => '_update_post_term_count',
+          'query_var'             => true,
+          'labels'                => array(
+            'name' => _x( 'Genres', 'taxonomy general name', 'textdomain' ),
+          ),
+
+        );
+
+        register_taxonomy( 'genres', array( 'post','page' ), $args );
+      });
+      """
+
+    When I run `wp taxonomy list --object_type=post --strict`
+    Then STDOUT should be a table containing rows:
+      | name        | label      | description | object_type | show_tagcloud | hierarchical | public |
+      | category    | Categories |             | post        | 1             | 1            | 1      |
+      | post_tag    | Tags       |             | post        | 1             |              | 1      |
+      | post_format | Formats    |             | post        |               |              | 1      |
+
+    When I run `wp taxonomy list --object_type=post --no-strict`
+    Then STDOUT should be a table containing rows:
+      | name        | label      | description | object_type | show_tagcloud | hierarchical | public |
+      | category    | Categories |             | post        | 1             | 1            | 1      |
+      | post_tag    | Tags       |             | post        | 1             |              | 1      |
+      | post_format | Formats    |             | post        |               |              | 1      |
+      | genres      | Genres     |             | post, page  | 1             | 1            | 1      |
--- a/features/taxonomy.feature
+++ b/features/taxonomy.feature
@@ -47,6 +47,7 @@ Feature: Manage WordPress taxonomies
       | name        | category   |
       | count       | 1          |
 
+  @require-wp-5.1
   Scenario: Listing taxonomies with strict/no-strict mode
     Given a WP installation
     And a wp-content/mu-plugins/test-taxonomy-list.php file:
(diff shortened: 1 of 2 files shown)


swissspidy:
What would strict mean here? What issue does it solve?

BhargavBhandari90:
@swissspidy I am working on this issue: https://github.com/wp-cli/entity-command/issues/181

src/Taxonomy_Command.php
```diff
@@ -148,12 +148,22 @@ protected function get_counts( $taxonomies ) {
 	public function list_( $args, $assoc_args ) {
 		$formatter = $this->get_formatter( $assoc_args );
 
+		// Check if it's strict mode or not.
+		$is_strict = ( isset( $assoc_args['strict'] ) && true === $assoc_args['strict'] ) ? true : false;
```
schlessera: There's a helper function called `get_flag_value` for retrieving flags like this. Providing a third argument of `false` means it defaults to `false`, i.e. `--no-strict`.

Also, by convention, we usually use the same name for the variable as we have for the flag: `--strict` => `$strict`.
```suggestion
		$strict = Utils\get_flag_value( $assoc_args, 'strict', false );
```

src/Taxonomy_Command.php
```diff
@@ -148,12 +148,22 @@ protected function get_counts( $taxonomies ) {
...
+			$taxonomy_object           = $assoc_args['object_type'];
+		} else {
+			$taxonomy_object = get_post_types();
 		}
 
 		$fields     = $formatter->fields;
-		$taxonomies = get_taxonomies( $assoc_args, 'objects' );
+		$taxonomies = ( $is_strict ) ?
+			get_taxonomies( $assoc_args, 'objects' ) :
+			get_object_taxonomies( $taxonomy_object, 'objects' );
```
schlessera: `$taxonomy_object` might not be defined here. So, further up we need to run a check to see if we have an object type if `--strict` was defined, and throw a `WP_CLI::error()` if not.

src/Taxonomy_Command.php
```diff
@@ -148,12 +148,22 @@ protected function get_counts( $taxonomies ) {
...
 		if ( isset( $assoc_args['object_type'] ) ) {
 			$assoc_args['object_type'] = array( $assoc_args['object_type'] );
+			$taxonomy_object           = $assoc_args['object_type'];
+		} else {
+			$taxonomy_object = get_post_types();
 		}
 
 		$fields     = $formatter->fields;
-		$taxonomies = get_taxonomies( $assoc_args, 'objects' );
+		$taxonomies = ( $is_strict ) ?
```
schlessera: No need for the parentheses.
Also, when using ternaries in multiple lines, it is usually preferable to put the ternary operators at the beginning of the line.
```suggestion
		$taxonomies = $strict
```

src/Taxonomy_Command.php
```diff
@@ -148,12 +148,22 @@ protected function get_counts( $taxonomies ) {
...
+			get_taxonomies( $assoc_args, 'objects' ) :
```
schlessera:
```suggestion
			? get_taxonomies( $assoc_args, 'objects' )
```

src/Taxonomy_Command.php
schlessera:
```suggestion
			: get_object_taxonomies( $taxonomy_object, 'objects' );
```

schlessera: > `$taxonomy_object` might not be defined here. So, further up we need to run a check to see if we have an object type if `--strict` was defined, and throw a `WP_CLI::error()` if not.

No, I need to correct myself. Throwing an error is not wanted here. What we want is to fall back to the previous default behavior if no object is provided. `get_object_taxonomies()` does not support this, so the entire condition should be reversed and extended:

```php
$taxonomies = ( isset( $taxonomy_object ) && ! $strict )
			? get_object_taxonomies( $taxonomy_object, 'objects' );
			: get_taxonomies( $assoc_args, 'objects' );
```

schlessera:
Great, code looks good. Can you add one or more functional tests to make sure it also works as expected?

BhargavBhandari90:
@schlessera . I'll be adding the functional tests over the weekend. Thanks 👍

features/taxonomy.feature
```diff
@@ -46,3 +46,42 @@ Feature: Manage WordPress taxonomies
...
       | count       | 1          |
+
+  Scenario: Listing taxonomies with strict/no-strict mode
+    Given a WP installation
+    And a wp-content/mu-plugins/test-taxonomy-list.php file:
+      """
+      <?php
+      // Plugin Name: Test Taxonomy Strict/No-Strict Mode
+
+      add_action( 'init', function(){
```
schlessera:
```suggestion
      add_action( 'init', function() {
```

features/taxonomy.feature
```diff
@@ -46,3 +46,42 @@ Feature: Manage WordPress taxonomies
...
+          'update_count_callback' => '_update_post_term_count',
+          'query_var'             => true,
+          'labels'                => array(
+            'name' => _x( 'Genres', 'taxonomy general name', 'textdomain' ),
+          ),
+
+        );
+
+        register_taxonomy( 'genres', array( 'post','page' ), $args );
+      });
```
schlessera:
```suggestion
      } );
```

thrijith:
@schlessera for [this](https://travis-ci.org/wp-cli/entity-command/jobs/531104195#L746), does it need a check here in [FeatureContext](https://github.com/wp-cli/wp-cli-tests/blob/83b636fc7cb69ff257e025d1c855c190143a45db/features/bootstrap/FeatureContext.php#L786) ?

schlessera:
Oh, thanks @thrijith, somehow I had looked at the wrong test output here... 🙃
I'll check what the issues are...

BhargavBhandari90:
@schlessera Somehow the test is failing. I am not sure about that. May be it's related to what @thrijith was saying...?

grappler:
Does the new flag not need to be documented in the function header?

schlessera:
@grappler Yes, indeed! That will be for the next release then, as I just pushed 2.6.0. :)

PR #273, opened 2019-11-04
Fix minor PHPCS violation

thrijith:
Fix violation causing build failure here https://travis-ci.org/wp-cli/entity-command/jobs/607188276#L587-L595

Got this when I merged https://github.com/wp-cli/entity-command/pull/266, not sure why it was not reported while `WP_CLI_CS` was implemented.

### DIFF
--- a/src/User_Session_Command.php
+++ b/src/User_Session_Command.php
@@ -176,7 +176,7 @@ class User_Session_Command extends WP_CLI_Command {
 
 		array_walk(
 			$sessions,
-			function( & $session, $token ) {
+			function( &$session, $token ) {
 				$session['token']           = $token;
 				$session['login_time']      = date( 'Y-m-d H:i:s', $session['login'] );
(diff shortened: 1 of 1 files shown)


thrijith:
@schlessera is the job failure due to the declaration of `feedback` in  https://github.com/wp-cli/wp-cli/blob/904f35f541697ac9c939c42b6742b99c32015d5b/php/WP_CLI/Compat/Min_PHP_5_6/FeedbackMethodTrait.php#L12  which accepts 2 params?

Also `$args_array` and `$arg` in the function doesn't seem to be used

```
public function feedback( $string, ...$args ) { // phpcs:ignore PHPCompatibility.LanguageConstructs.NewLanguageConstructs.t_ellipsisFound
	$args_array = [];
	foreach ( $args as $arg ) {
		$args_array[] = $args;
	}

	$this->process_feedback( $string, $args );
}
```

PR #275, opened 2019-11-06
Recalculate menu order on insertion or deletion

petruchek:
New private method called reorder_menu_items() which is being called after any menu item is being deleted or new menu item is being added and it has explicitly set position.

Issue reported and briefly discussed here: https://github.com/wp-cli/wp-cli/issues/5047

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,10 @@
     },
     "config": {
         "process-timeout": 7200,
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "dealerdirect/phpcodesniffer-composer-installer": true
+        }
     },
     "extra": {
         "branch-alias": {
--- a/features/menu-item.feature
+++ b/features/menu-item.feature
@@ -34,10 +34,10 @@ Feature: Manage WordPress menu items
     When I run `wp menu item add-term sidebar-menu post_tag {TERM_ID} --porcelain`
     Then save STDOUT as {TERM_ITEM_ID}
 
-    When I run `wp menu item add-custom sidebar-menu Apple http://apple.com --parent-id={POST_ITEM_ID} --porcelain`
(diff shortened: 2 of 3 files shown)


petruchek:
Hi there @schlessera

Any news on this PR being merged?

Thanks

schlessera:
Hi @petruchek,

Yes, I'd like to move this forward. However, we'll need Behat tests for this first. Can you add tests to assert that the menu IDs are correctly modified?

Let me know if you need assistance with writing the tests!

petruchek:
> Let me know if you need assistance with writing the tests!

@schlessera, thank you for your response. 

Could you please guide me a little, please?

I'm assuming I need to modify features/menu-item.feature. Is this correct?
Do you think I need to create a separate scenario? Or I should modify existing scenarios?

Thank you for your time (very much)

schlessera:
@petruchek 

> I'm assuming I need to modify features/menu-item.feature. Is this correct?

Yes, that's the one to edit.

> Do you think I need to create a separate scenario? Or I should modify existing scenarios?

Please create a separate scenario. In the existing tests, you can already see how the position is being represented in the output lists, so it should be relatively straight-forward to copy one of the existing scenarios and adapt it to test this PR.

petruchek:
> @petruchek
> 
> > I'm assuming I need to modify features/menu-item.feature. Is this correct?
> 
> Yes, that's the one to edit.
> 
> > Do you think I need to create a separate scenario? Or I should modify existing scenarios?
> 
> Please create a separate scenario. In the existing tests, you can already see how the position is being represented in the output lists, so it should be relatively straight-forward to copy one of the existing scenarios and adapt it to test this PR.

Hi there,

Is it something that still makes sense doing? Please let me know.

schlessera:
Hey @petruchek,
Yes, this still makes sense.

schlessera:
Thanks for the PR, @petruchek !

PR #284, opened 2020-03-07
Update `site empty` command to include removal of `wp_links` table data

### LINKED ISSUE #261: The links table should also be emptied
When running `wp site empty` the `links` table does not get emptied. It should do, as it contains user content in the same way that the posts, comments, terms, and meta tables do.

thrijith:
Fixes #261

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -116,6 +116,34 @@ class Site_Command extends CommandWithDBObject {
 	}
 
 	/**
+	 * Delete all links, link_category terms, and related cache.
+	 */
+	private function empty_links() {
+		global $wpdb;
+
+		// Remove links and related cached data.
+		$links_query = "SELECT link_id FROM $wpdb->links";
+		$links       = new QueryIterator( $links_query, 10000 );
+
+		// Remove bookmarks cache group.
+		wp_cache_delete( 'get_bookmarks', 'bookmark' );
+
+		while ( $links->valid() ) {
+			$link_id = $links->current()->link_id;
+
+			// Remove cache for the link.
+			wp_delete_object_term_relationships( $link_id, 'link_category' );
+			wp_cache_delete( $link_id, 'bookmark' );
+			clean_object_term_cache( $link_id, 'link' );
+
+			$links->next();
+		}
+
+		// Empty the table once link related cache and term is removed.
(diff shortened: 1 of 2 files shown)


thrijith:
@schlessera I have manually tested and verified removal of data on local setup, I am not sure how to go about adding tests for the same, please let me know if a test is required for this change. Thanks.

schlessera:
There are already existing tests for the `site empty` command here: https://github.com/wp-cli/entity-command/blob/v2.0.7/features/site-empty.feature

You can just add to these to verify the logic with the following test:
- add an entry to the `wp_links` table;
- run `site empty`;
- query the `SELECT COUNT(*) from $wpdb->links` and verify that the count is zero.

src/Site_Command.php
```diff
@@ -115,6 +115,34 @@ private function empty_taxonomies() {
...
 	}
 
+	/**
+	 * Delete all links, link_category terms, and related cache.
+	 */
+	private function empty_links() {
+		global $wpdb;
+
+		// Remove links and related cached data.
+		$links_query = "SELECT link_id FROM $wpdb->links";
```
schlessera: Convention to always use curly braces on text expansion:
```suggestion
		$links_query = "SELECT link_id FROM {$wpdb->links}";
```

src/Site_Command.php
```diff
@@ -115,6 +115,34 @@ private function empty_taxonomies() {
...
+			// Remove cache for the link.
+			wp_delete_object_term_relationships( $link_id, 'link_category' );
+			wp_cache_delete( $link_id, 'bookmark' );
+			clean_object_term_cache( $link_id, 'link' );
+
+			$links->next();
+		}
+
+		// Empty the table once link related cache and term is removed.
+		$wpdb->query( "TRUNCATE $wpdb->links" );
```
schlessera:
```suggestion
		$wpdb->query( "TRUNCATE {$wpdb->links}" );
```

thrijith:
@schlessera the build seems to have passed, but there seems to be some issue with config, can you please take look when time permits? Thanks.

If the travis config isn't valid, maybe I can create a PR to update that as well, seems like a good option to use automated script to add updates across repos.

schlessera:
@thrijith I'm pushing PRs to all repos right now to fix the Travis warning.

PR #286, opened 2020-03-25
Fix Travis CI `sudo` warning

schlessera:
Travis currently shows a warning that the configuration option `sudo` has been deprecated.

As there's another information (not a warning yet) that the configuration argument `os` is missing and defaults to `linux`, this PR replace `sudo: false` with `os: linux`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+os: linux
 dist: trusty
 
 language: php


thrijith:
We don't have an open milestone, @schlessera  should it be `2.0.8` or be in sync with minor version of the framework?

schlessera:
For command packages, we always just create the next milestone, using semver to decide whether it's a patch, minor or major release.

PR #289, opened 2020-04-15
Add wp_slash() option to the check-password subcommand

fjarrett:
Partially addresses https://github.com/wp-cli/wp-cli/issues/5278

Should this be saved for WP-CLI 3.0? Potentially, yes.

This PR is done in a way that is backward compatible, so that manual escaping is still the norm and you can use a new flag to have WP-CLI escape the characters for you.

But IMHO automatic escaping via `wp_slash()` should be the default behavior, after all, this is how the `wp-login.php` POST works, and as described in https://github.com/wp-cli/wp-cli/issues/5278, the `wp user create` command doesn't require you to escape the password characters, and it stands to reason that you should be able to check the password using the exact same input.

Perhaps in the next major release we could change the flag value be `true` by default so users can simply do `--no-escape-chars` to get back the current 2.x behavior and escape everything manually.

**CURRENT BEHAVIOR**
```bash
wp user create admin admin@local.test --user_pass="admin'spass"
wp user check-password admin "admin'spass"
echo $?
1

wp user create admin admin@local.test --user_pass="admin'spass"
wp user check-password admin "admin\'spass"
echo $?
0
```

**NEW BEHAVIOR**
```bash
wp user create admin admin@local.test --user_pass="admin'spass"
wp user check-password admin "admin'spass"
Warning: Password contains characters that need to be escaped. Please escape them manually or use the `--escape-chars` option.
echo $?
1

wp user create admin admin@local.test --user_pass="admin'spass"
wp user check-password admin "admin'spass" --escape-chars
echo $?
0
```

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -85,6 +85,28 @@ Feature: Manage WordPress users
       """
     And the return code should be 1
 
+    When I run `wp user create testuser3b testuser3b@example.com --user_pass="test\"user3b's\pass\!"`
+    Then STDOUT should not contain:
+       """
+       Password:
+       """
+
+    # Check password without the `--escape-chars` option.
+    When I try `wp user check-password testuser3b "test\"user3b's\pass\!"`
+    Then STDERR should be:
+      """
+      Warning: Password contains characters that need to be escaped. Please escape them manually or use the `--escape-chars` option.
+      """
+    And the return code should be 1
+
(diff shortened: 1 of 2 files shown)


Basilakis:
@schlessera That is not on the Documentation, it should be added there :) The --escape-chars

PR #304, opened 2020-11-25
Better error message when trying to delete super admin

tyrann0us:
WP-CLI currently does not provide information on why the deletion of a user failed.

A common reason for failures is when trying to delete a Super Admin ([`wpmu_delete_user()`](https://developer.wordpress.org/reference/functions/wpmu_delete_user/) prevents this from happening, see https://github.com/WordPress/WordPress/blob/e89f6692745937af4bfe5acbf8c5918f393bbc77/wp-admin/includes/ms.php#L153-L157).

This PR tries to improve the error message by extending `Warning: Failed deleting user {$user_id}.` with `The user is a Super Admin.`, if applicable.

Note that this is untested; all changes have been made with the GitHub UI.


tyrann0us:
Looks like Behat tests fail with `Error: 'super-admin' is not a registered wp command.`. I have no idea what's going on.

tyrann0us:
Looks like the code works; the error message is correctly extended. But somehow Behat doesn't like the exit code `1` of the of the intentionally failing WP-CLI command, despite being told to expect it. 🤔

tyrann0us:
@schlessera, awesome job to migrate from Travis to GitHub Actions! It was very annoying to wait at least 45 mins before the tests started.

Any idea why the WordPress `v3.7.11` test runs for over an hour now already? All other finished in ~6 minutes. Oddly, GitHub Actions doesn't let me see details of the running "Run Behat" step, as it normally would.

composer.json
```diff
@@ -128,6 +129,10 @@
             "site unarchive",
             "site unmature",
             "site unspam",
+            "super-admin",
+            "super-admin add",
+            "super-admin list",
+            "super-admin remove",
```
schlessera, line 135: These are not provided by this repository, so shouldn't appear here.
```suggestion
```

schlessera (changes requested):

schlessera (approved):

PR #326, opened 2021-08-18
Update example input to reflect example output

thomasplevy:
This is (I know) a small documentation update that hardly matters *but* while looking at the command `wp post exists --help` I noticed the small inconsistency and figured I'd submit a quick patch instead of ignoring it.

Thanks <3


schlessera (approved):

thomasplevy:
:rocket:

PR #346, opened 2022-02-10
Add --single option to CommandWithMeta::get method

humanik:
Add --single option to CommandWithMeta::get method


danielbachhuber (commented):
Thanks for the pull request, @humanik !

Can you include a Behat feature test for this change?

Issue #34, opened 2017-07-24
Allow filtering user list by users with no role


johnbillion:
The `wp user list` command accepts a `--role` filter, but this does not support users with no role.

In wp-admin, the Users screen now supports filtering by users with no role. It uses `none` as the role name.

Bear in mind there is a bug in core at the moment which prevents such users from appearing when multisite is enabled. Might want to wait until this is fixed: https://core.trac.wordpress.org/ticket/36196

johnbillion:
The core bug is now fixed in trunk, scheduled for 4.9.

danielbachhuber:
@johnbillion Based on the commit, it seems like WP-CLI is unaffected and could implement this enhancement without WP 4.9 as a requirement. Is my assessment correct?

> It uses none as the `role` name.

To clarify, the WP-CLI implementation would be `wp user list --role=none` ?

johnbillion:
I'll take a look.

danielbachhuber:
@johnbillion Can you write a simple Behat test to describe how you'd like this to behave?

BhargavBhandari90:
I think for this, we need to allow user adding without role using `wp user create`?

OR we are just looking for getting users with no role?

Currently adding PR for getting users with no role only.

danielbachhuber:
> I think for this, we need to allow user adding without role using `wp user create`?
> 
> OR we are just looking for getting users with no role?

Just the latter at the moment — users with no role.

PR #360, opened 2022-08-14
Support filtering for users without a role with `--role=none`

BhargavBhandari90:
Fixes https://github.com/wp-cli/entity-command/issues/34

Allow getting users with no role by using the following command

`wp user list --role=none`

Waiting for clarification https://github.com/wp-cli/entity-command/issues/34#issuecomment-1214336137

### DIFF
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -162,7 +162,17 @@ class User_Command extends CommandWithDBObject {
 
 		$assoc_args['count_total'] = false;
 		$assoc_args                = self::process_csv_arguments_to_arrays( $assoc_args );
-		$users                     = get_users( $assoc_args );
+
+		if ( ! empty( $assoc_args['role'] ) && 'none' === $assoc_args['role'] ) {
+			$norole_user_ids = wp_get_users_with_no_role();
+
+			if ( ! empty( $norole_user_ids ) ) {
+				$assoc_args['include'] = $norole_user_ids;
+				unset($assoc_args['role']);
+			}
+		}
+
+		$users = get_users( $assoc_args );
 
 		if ( 'ids' === $formatter->format ) {
 			echo implode( ' ', $users );
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -168,7 +168,7 @@ class User_Command extends CommandWithDBObject {
 
 			if ( ! empty( $norole_user_ids ) ) {
 				$assoc_args['include'] = $norole_user_ids;
-				unset($assoc_args['role']);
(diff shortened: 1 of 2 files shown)


src/User_Command.php
```diff
@@ -162,7 +162,17 @@ public function list_( $args, $assoc_args ) {
 
 		$assoc_args['count_total'] = false;
 		$assoc_args                = self::process_csv_arguments_to_arrays( $assoc_args );
-		$users                     = get_users( $assoc_args );
+
+		if ( ! empty( $assoc_args['role'] ) && 'none' === $assoc_args['role'] ) {
```
danielbachhuber, line 166: @wp-cli/committers Do you think this should be `--role=none` or `--no-role`? I think I'm leaning towards the latter...

danielbachhuber (changes requested):
Thanks for kicking this off, @BhargavBhandari90 !

Can you add some functional tests (Behat) for this pull request?

src/User_Command.php
BhargavBhandari90: I was following this https://github.com/wp-cli/entity-command/issues/34#issuecomment-317870209

We can change it to whatever you say. Just let me know. :)

When you say `--no-role` means the command should be wp user list `--no-role` ?

BhargavBhandari90:
> Thanks for kicking this off, @BhargavBhandari90 !
> 
> Can you add some functional tests (Behat) for this pull request?

Sure, I'll add the behat test

src/User_Command.php
danielbachhuber: > When you say `--no-role` means the command should be `wp user list --no-role` ?

Yep, exactly.

> We can change it to whatever you say. Just let me know. :)

@johnbillion @schlessera Do either of you have an opinion?

schlessera: I think `--no-role` might be misleading, as this is typically used for boolean flags, and would make me assume `--role` (without any value) is a thing, too.
I'd prefer `--role=none` here to make it clear that this is not a boolean flag.

johnbillion: +1 for `--role=none` which matches the URL used in wp-admin: `?role=none`.

BhargavBhandari90: @johnbillion is right.

Thank You all guys. So we'll keep `--role=none`

I'll start adding behat test now

BhargavBhandari90: I think we can Resolve this conversation. Let me know.

Thanks

danielbachhuber (approved):

PR #363, opened 2022-08-21
WIP: Issue #87 : Show primary site of the user

### LINKED ISSUE #87: Add a positional argument to get primary site in multisite
As of now we don't have any means to get the primary site using WP-CLI
We can have a command as `wp site list --primary` to get the name and ID of the primary site that is set at Dashboard > My Sites

BhargavBhandari90:
Fixes #87 

@Sidsector9 If there are multiple administrators on the site, they can set different primary sites for them.

image

In that case, we have to pass `--user` in this command.

Let me know if this approach works. Then I'll add behat tests.

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -574,6 +574,16 @@ class Site_Command extends CommandWithDBObject {
 
 		global $wpdb;
 
+		if ( isset( $assoc_args['primary'] ) ) {
+			$user_id = get_current_user_id();
+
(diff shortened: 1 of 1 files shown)


danielbachhuber (changes requested):
I'm not quite sure this is the approach we want to take, see https://github.com/wp-cli/entity-command/issues/87#issuecomment-1222927945

BhargavBhandari90:
Closing this as no longer relevent

PR #371, opened 2022-09-11
Show error when parameter is missing in wp option patch

### LINKED ISSUE #353: Missing argument with `option patch update` should throw error
Hello, I'm having problems using this commands to disable and enable the maintenance mode of a subsite.
If the maintenance mode is enabled and i send the command to disable it the result is:
wp --url=subsite.multisite.com --allow-root option patch update seedprod_settings '{"enable_maintenance_mode":"false"}' --format=json
Success: Updated 'seedprod_settings' option.

If i send the command to enable the maintenance mode again the result is:
wp --url=subsite.multisite.com --allow-root option patch update seedprod_settings '{"enable_maintenance_mode":"true"}' --format=json
PHP Warning:  json_decode() expects parameter 1 to be string, array given in /home/multisite.com/public_html/wp-content/plugins/coming-soon/app/render-csp-mm.php on line 36
array (
  'enable_maintenance_mode' => 'true',
)

And the seedprod_settings object settings dissapear from the subsite options. Any idea what can be happening here?
Thanks in advance for the support.

BhargavBhandari90:
Fixes #353 

- Currently, I've displayed an error message when an argument is missing.

### DIFF
--- a/features/option-pluck-patch.feature
+++ b/features/option-pluck-patch.feature
@@ -317,3 +317,10 @@ Feature: Option commands have pluck and patch.
       Please provide value to update.
       """
     And the return code should be 1
+
+    When I try `wp option patch update option_name foo 0`
+    And STDERR should contain:
+      """
+      Please provide value to update.
+      """
+    And the return code should be 1
--- a/features/option-pluck-patch.feature
+++ b/features/option-pluck-patch.feature
@@ -299,3 +299,21 @@ Feature: Option commands have pluck and patch.
       """
       1
       """
+
+  @patch
+  Scenario: When we don't pass all necessary argumants.
+    Given a WP install
+    And an input.json file:
+      """
+      {
+        "foo": "bar"
+      }
+      """
+    And I run `wp option update option_name --format=json < input.json`
+
+    When I try `wp option patch update option_name foo`
+    And STDERR should contain:
+      """
+      Please provide value to update.
+      """
+    And the return code should be 1
--- a/features/option-pluck-patch.feature
+++ b/features/option-pluck-patch.feature
@@ -318,9 +318,8 @@ Feature: Option commands have pluck and patch.
       """
     And the return code should be 1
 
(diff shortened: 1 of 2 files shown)


src/Option_Command.php
```diff
@@ -569,7 +569,13 @@ function( $key ) {
 				: null;
 			$patch_value = ! empty( $stdin_value )
 				? WP_CLI::read_value( $stdin_value, $assoc_args )
-				: WP_CLI::read_value( array_pop( $key_path ), $assoc_args );
+				: ( count( $key_path ) > 1
+					? WP_CLI::read_value( array_pop( $key_path ), $assoc_args )
+					: null );
```
danielbachhuber, line 574: This nested ternary is a bit complex to read 😕

Can we expand it to some if/else statements?

features/option-pluck-patch.feature
```diff
@@ -299,3 +299,21 @@ Feature: Option commands have pluck and patch.
...
+      }
+      """
+    And I run `wp option update option_name --format=json < input.json`
+
+    When I try `wp option patch update option_name foo`
+    And STDERR should contain:
+      """
+      Please provide value to update.
+      """
+    And the return code should be 1
```
danielbachhuber, line 319: Can we add a test case for:

```
When I run `wp option patch update option_name foo 0`
```

To make sure we're handling falsy values correctly?

BhargavBhandari90: It should not allow this to update, right?

danielbachhuber: > It should not allow this to update, right?

Perhaps I'm misunderstanding but isn't `0` a valid value?

BhargavBhandari90: Actually, I misunderstood.

I thought zero should be considered an empty value.

Then I saw other commands in the same `option-pluck-patch.feature` file.

So fixe it.

danielbachhuber (approved):
Thanks for the updates!

PR #372, opened 2022-09-11
Allow `user_url` to be set with `wp user create`

### LINKED ISSUE #350: `wp user create` ignores the `user_url` argument
Hi,

When using command line `wp create user` with --user_url argument, the value is ignored.

Step to reproduce : 

`/usr/bin/php /usr/local/bin/wp-cli --path='/path/to/wordpress/' user create 'testurl' 'sample@email.com' --role='editor' --user_pass='strong_password' --first_name='first' --last_name='last' --user_url='http://www.testsite.com' --debug`

No error reported in debug.

Then :

`/usr/local/bin/wp-cli --path='/path/to/wordpress/' user get testurl --fields=user_url`

Return empty value instead of http://www.testsite.com

kjohnson:
Fixes #350 by setting `$user->user_url` from the `user_url` flag value.

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -497,3 +497,12 @@ Feature: Manage WordPress users
       Success: Updated user 1.
       """
     And an email should not be sent
+
+  Scenario: Set user url when creating a user
+    Given a WP install
+    And I run `wp user create testurl sample@email.com --user_url='http://www.testsite.com'`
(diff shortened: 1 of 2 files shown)


danielbachhuber (approved):
🎉 

Great work!

PR #375, opened 2022-09-25
Issue #374 : Add examples for wp option patch command

### LINKED ISSUE #374: Add examples for `wp option patch`
- We don't have examples of commands for `wp option patch` at [wp option patch](https://developer.wordpress.org/cli/commands/option/patch/)
- So we should add that

--------
NOTE: There are many others, I am starting with this to get an idea of what examples we should add.

BhargavBhandari90:
Fixes #374 

- Added examples for `wp option patch`

### DIFF
--- a/src/Option_Command.php
+++ b/src/Option_Command.php
@@ -555,7 +555,7 @@ class Option_Command extends WP_CLI_Command {
 	 *     $ wp option patch insert option_name foo bar
 	 *     Success: Updated 'option_name' option.
 	 *
-	 *     # Patch Update.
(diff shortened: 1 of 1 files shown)


danielbachhuber (commented):
Thanks, @BhargavBhandari90 !

Could we be a little more descriptive with the commands? I've provided a couple examples to get you started, but I think they could all be improved.

BhargavBhandari90:
@danielbachhuber
Thanks for the suggestion.
I've updated the notes in the examples.
Let me know if that works.

danielbachhuber (approved):
Looks great! Thanks for your work on this, @BhargavBhandari90

PR #378, opened 2022-10-01
Use `current_time()` for setting the default value for `user_registered` in `wp user create`

### LINKED ISSUE #377: `wp user create` with no `--user_registered` arg creates users with invalid registration date
## Bug Report

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [X] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

(Discovered in https://github.com/woocommerce/woocommerce/issues/34894).

When creating users via `wp user create` and not specifying a value for the optional field `--user_registered`, the user ends up with "0000-00-00 00:00:00" as the value of the `user_registered` column in the database (i.e. the default value for this column in the `wp_users` table).

According to the docs, this field should default to the current date/time:
https://github.com/wp-cli/entity-command/blob/59c591e021c5c528c91bca95e370dfc7f9ca1a90/src/User_Command.php#L338-L339

This relates to commit a6acb64d1276baeb8a9f2f391a3a7d5cd4a6b039, in which the default value for `user_registered` was set to the result of a `date_format()` call instead of the previous `strftime()` call, without changing the placeholders in the format string (`'%F %T'`):

- In [`strftime()`](https://www.php.net/manual/en/function.strftime) `%F %T` produced the correct result because `%F` is the same as `%Y-%m-%d` and `%T` is a shorthand for `%H:%M:%S`.
- In [`date_format()`](https://www.php.net/manual/en/datetime.format.php) not only the `%` symbol is not used, but `F` is the full textual representation of a month and `T` the timezone abbreviation. This means users are being created with `user_registered` set to something like "%September %UTC", which is incorrect.

- Run `wp user create testuser testuser@example.com`.
- Check that `wp user get testuser` shows `0000-00-00 00:00:00` as the value for the `user_registered` field.

Per the docs, the user should've been created with the current date/time as `user_registered`.

```
OS:	Darwin 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64
Shell:	/bin/zsh
PHP binary:	/opt/homebrew/bin/php
PHP version:	8.1.11
php.ini used:	/opt/homebrew/etc/php/8.1/php.ini
MySQL binary:	/opt/homebrew/bin/mysql
MySQL version:	mysql  Ver 14.14 Distrib 5.7.39, for osx10.17 (x86_64) using  EditLine wrapper
SQL modes:	
WP-CLI root dir:	/Users/jorge/src/wpcli-entity-command/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	/Users/jorge/src/wpcli-entity-command/vendor
WP_CLI phar path:	
WP-CLI packages dir:	
WP-CLI global config:	
WP-CLI project config:	/Users/jorge/src/wpcli-entity-command/wp-cli.yml
WP-CLI version:	2.6.0
```

Use the proper placeholders in the date `$format` string in `date_format()` or rely on [`current_time( 'mysql', true )`](http://developer.wordpress.org/reference/functions/current_time/) which already produces a string in the desired format.

jorgeatorres:
This PR fixes issue #377, by relying on `current_time()` instead of `date_format()` to properly produce a default value for `user_registered` when creating users via `wp user create`.

I've also included a behat test that should fail in `master` (but not on this branch).

**Note:** I did a quick search through the org and this seems to be the only place where we were using `date_format()` with an invalid format.

### Steps to test

1. Check out `master`.
1. Run `wp user create testuser testuser@example.com`.
1. Check that `wp user get testuser` shows `0000-00-00 00:00:00` as the value for the `user_registered` field.
1. Delete the user created in step 2. For example, with `wp user delete testuser`.
1. Check out this branch.
1. Repeat step 2 and confirm that this time the current date/time is the value of the `user_registered` field.

Closes #377.

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -17,6 +17,12 @@ Feature: Manage WordPress users
       | ID           | {USER_ID}  |
       | roles        | author     |
 
+    When I run `wp user get {USER_ID} --field=user_registered`
+    Then STDOUT should not contain:
+      """
+      0000-00-00 00:00:00
+      """
+
     When I run `wp user meta get {USER_ID} first_name`
     Then STDOUT should be:
       """
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -397,7 +397,7 @@ class User_Command extends CommandWithDBObject {
 		$user->user_registered = Utils\get_flag_value(
 			$assoc_args,
 			'user_registered',
-			date_format( date_create(), '%F %T' )
+			current_time( 'mysql', true )
 		);
 
 		$user->display_name = Utils\get_flag_value( $assoc_args, 'display_name', false );
@@ -978,7 +978,7 @@ class User_Command extends CommandWithDBObject {
 			$defaults = [
 				'role'            => get_option( 'default_role' ),
 				'user_pass'       => wp_generate_password(),
-				'user_registered' => date_format( date_create(), '%F %T' ),
+				'user_registered' => current_time( 'mysql', true ),
 				'display_name'    => false,
 			];
 			$new_user = array_merge( $defaults, $new_user );


danielbachhuber (approved):
Looks great, thanks @jorgeatorres ! I appreciate your detailed description 😊

PR #382, opened 2022-11-01
Add `wp option get-autoload` and `wp option set-autoload`

danielbachhuber:
Adds `wp option get-autoload` and `wp option set-autoload` commands:

```
$ wp option add foo bar
Success: Added 'foo' option.
$ wp option get-autoload foo
yes
$ wp option set-autoload foo no
Success: Updated autoload value for 'foo' option.
$ wp option set-autoload foo no
Success: Autoload value passed for 'foo' option is unchanged.
```

These are useful for manipulating the `autoload` value. `wp option update` can't be used for this directly because `update_option()` won't change `autoload` if the `value` doesn't change.

Recreates the cache manipulation inside of `update_option()` too: https://github.com/WordPress/wordpress-develop/blob/04e8bb4bb5ed37dade1b1ddda634e45f7503820d/src/wp-includes/option.php#L525-L540

This feels brittle, but not sure there's a better way around it...

### DIFF
--- a/src/Option_Command.php
+++ b/src/Option_Command.php
@@ -473,7 +473,7 @@ class Option_Command extends WP_CLI_Command {
 	 * : The name of the option to set 'autoload' for.
 	 *
 	 * <autoload>
-	 * : Requires WP 4.2. Should this option be automatically loaded.
+	 * : Should this option be automatically loaded.
 	 * ---
 	 * options:
 	 *   - 'yes'
--- a/src/Option_Command.php
+++ b/src/Option_Command.php
@@ -437,6 +437,98 @@ class Option_Command extends WP_CLI_Command {
 	}
 
 	/**
+	 * Gets the 'autoload' value for an option.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <key>
+	 * : The name of the option to get 'autoload' of.
+	 *
+	 * @subcommand get-autoload
+	 */
+	public function get_autoload( $args ) {
+		global $wpdb;
+
+		list( $option ) = $args;
+
+		$existing = $wpdb->get_row(
+			$wpdb->prepare(
+				"SELECT autoload FROM $wpdb->options WHERE option_name=%s",
+				$option
+			)
+		);
+		if ( ! $existing ) {
+			WP_CLI::error( "Could not get '{$option}' option. Does it exist?" );;
+		}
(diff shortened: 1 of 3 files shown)


src/Option_Command.php
```diff
@@ -436,6 +436,100 @@ public function update( $args, $assoc_args ) {
...
+	/**
+	 * Sets the 'autoload' value for an option.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <key>
+	 * : The name of the option to set 'autoload' for.
+	 *
+	 * <autoload>
+	 * : Requires WP 4.2. Should this option be automatically loaded.
```
schlessera, line 476: Why would this require 4.2? As far as I understand, 4.2 added the `$autoload` arg to `update_option()`, but we're not using that here. Given the `autoload` DB field was introduced with 3.4 (or even earlier?), this should work with all WP version supported by WP-CLI, shouldn't it?

src/Option_Command.php
```diff
@@ -436,6 +436,100 @@ public function update( $args, $assoc_args ) {
...
+	 * ---
+	 * options:
+	 *   - 'yes'
+	 *   - 'no'
```
schlessera, line 480: This means the API is:
```
wp option set-autoload my_option no
```
That seems to be forced by the DB value and is neither developer-speak (set autoload true) nor natural language (disable autoload).

I'm don't love that API, but I have trouble coming up with something better.

What do think about:
```
wp option autoload get my_option
wp option autoload enable my_option
wp option autoload disable my_option
```

schlessera (changes requested):

src/Option_Command.php
danielbachhuber:
```suggestion
	 * : Should this option be automatically loaded.
```

@schlessera You are correct! I originally started to use `update_option()`, hence the comment, but had to write my own implementation because `update_option()` bails early if the value isn't changing.

src/Option_Command.php
danielbachhuber: > I'm don't love that API, but I have trouble coming up with something better.
> 
> What do think about:

@schlessera Good points. I don't love your proposed approach either because it seems unnecessarily complex to nest the commands further.

I went with `wp option set-autoload my_option no` because `wp option add` accepts `--autoload=yes`/`--autoload=no`.

PR #387, opened 2022-12-13
add example for deleting multiple post IDs

### LINKED ISSUE #386: clarify syntax to list multiple post IDs to delete
Examples of how to delete posts only include a single ID deletion: 
[https://developer.wordpress.org/cli/commands/post/update/#examples](url)
leaving it unclear how multiple IDs should be listed.

Other example sections [https://developer.wordpress.org/cli/commands/post/update/#examples](url) make it clear that the list should be a space separated, not comma separated.

joeldcanfield:
The documentation isn't clear that to delete multiple posts the IDs must be space separated, not comma separated

Fixes #386

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -469,6 +469,12 @@ class Post_Command extends CommandWithDBObject {
 	 *     $ wp post delete 123 --force
 	 *     Success: Deleted post 123.
 	 *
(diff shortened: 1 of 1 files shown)


danielbachhuber (approved):
Looks great, thanks!

PR #393, opened 2023-02-27
Do not mark main site as spam when using 'wp user spam'

r-a-y:
If a user has a role on the main site and we use `wp user spam` on that user, the main site will be marked as spam.

It looks like the following commit broke the main site check: https://github.com/wp-cli/entity-command/commit/60f91b3f3ce7a3109ddcbc799ebe3b6a1e574237. Commit 1bc55d0 fixes this by using `is_main_site()`.

I've also removed the call to `use WP_CLI\Fetchers\Site as SiteFetcher` in commit 5f4b7e1 since it doesn't appear to be necessary.

I haven't written a behat test for this yet, but feel free to do so.

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -435,6 +435,15 @@ Feature: Manage WordPress users
     And I run `wp user get oprime`
     Then STDOUT should not be empty
 
+    When I run `wp site create --slug=foo --porcelain`
+    Then save STDOUT as {SPAM_SITE_ID}
+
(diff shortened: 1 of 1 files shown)


danielbachhuber (approved):
Thanks for the PR, @r-a-y !

It looks like the original incorrect logic is from https://github.com/wp-cli/entity-command/pull/74

I verified that WordPress core's behavior in wp-admin is to:
1. Spam any of the user's blogs.
2. Not spam the main site.

I also added some Behat tests to cover the changes

Issue #342, opened 2022-01-26
"user reset-password" doesn't display the new randomly-generated password, suggest adding a switch to enable


dsXLII:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

`wp user reset-password username` does not display the new password. 

As a comparison/example: when you create a new user with the CLI, and don't specify a password, the new randomly-generated password is displayed. Admittedly this is a different use case, in that you only create one account at a time with `wp user create` but you can reset multiple users' passwords.

**Describe the solution you'd like**

When resetting the password of a single user, I would like the option of having the password displayed. Something like `wp user reset-password foo --show-password` or similar.

This probably should NOT be done by default, because that would be a change to existing behavior, and if users are currently doing password resets in an automated fashion could lead to passwords being logged in places where they weren't before and thus comes as a surprise. And I'm not sure it makes sense to do so when multiple users are specified on the command line. (Or at least that's a bit more of a refactor than adding a switch, as you'd then have to handle fields and otherwise do a proper "generate output" routine.)

danielbachhuber:
I'm open to a `--show-password` flag if someone wants to put together a pull request for it.

dsXLII:
Completely forgot about this, this was one whole job ago. :) 

Anyway, https://github.com/wp-cli/entity-command/pull/394 .

dsXLII:
And a possible opportunity for improvement: When you create a new user, we call `wp_generate_password(24)` but here we only call `wp_generate_password()` which defaults to a 12-character password. Any specific reason for the discrepancy?

danielbachhuber:
🚢

PR #394, opened 2023-03-25
Add --show-password flag to user reset-password

dsXLII:
First go at a PR to address https://github.com/wp-cli/entity-command/issues/342 .

### DIFF
--- a/features/user-reset-password.feature
+++ b/features/user-reset-password.feature
@@ -61,3 +61,20 @@ Feature: Reset passwords for one or more WordPress users.
     """
     {ORIGINAL_PASSWORD}
     """
+
+  @require-wp-4.3
+  Scenario: Reset the password of a WordPress user, and show only the new password
+    Given a WP installation
+
+    When I run `wp user get 1 --field=user_pass`
+    Then save STDOUT as {ORIGINAL_PASSWORD}
+
+    When I run `wp user reset-password 1 --skip-email --porcelain`
+    Then STDOUT should not be empty
+    And an email should not be sent
+
+  When I run `wp user get 1 --field=user_pass`
+  Then STDOUT should not contain:
+    """
+    {ORIGINAL_PASSWORD}
+    """
--- a/features/user-reset-password.feature
+++ b/features/user-reset-password.feature
@@ -56,11 +56,11 @@ Feature: Reset passwords for one or more WordPress users.
       """
     And an email should not be sent
 
-  When I run `wp user get 1 --field=user_pass`
-  Then STDOUT should not contain:
-    """
-    {ORIGINAL_PASSWORD}
-    """
+    When I run `wp user get 1 --field=user_pass`
+    Then STDOUT should not contain:
+      """
+      {ORIGINAL_PASSWORD}
+      """
 
   @require-wp-4.3
   Scenario: Reset the password of a WordPress user, and show only the new password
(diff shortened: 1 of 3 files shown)


features/user-reset-password.feature
```diff
@@ -41,3 +41,23 @@ Feature: Reset passwords for one or more WordPress users.
...
       """
+
+@require-wp-4.3
+Scenario: Reset the password of a WordPress user, and show the new password
+  Given a WP installation
+
+  When I run `wp user get 1 --field=user_pass`
+  Then save STDOUT as {ORIGINAL_PASSWORD}
+
+  When I run `wp user reset-password 1 --skip-email --show-password`
```
dsXLII, line 52: This absolutely should capture the new password, and the next step should check for it, but I have no idea how to do that :(

danielbachhuber (commented):
Great start!

danielbachhuber:
@dsXLII Still planning to finish this up?

dsXLII:
I'm still here, just distracted by the day job :/

Question about porcelain. For the new app password example, since you can only create one app password at a time the output is unambiguous. But here, since you can reset multiple users' passwords at the same time, it's less clear. (If I do a `wp user reset-password 1 2 3 4` four passwords would be output, but it's not obvious that the passwords go in the same order as the user arguments.) What's the best way to handle that? Output them in that order, and document it? Don't allow --porcelain for multiple user arguments?

danielbachhuber:
> I'm still here, just distracted by the day job :/

No worries! 

> Question about porcelain.

Good question! You are correct: we don't yet have prior art within WP-CLI.

Let's follow `git status` and output each password on a new line:

```
$ git status --porcelain
?? test1
?? test2
```

features/user-reset-password.feature
```diff
@@ -41,3 +41,40 @@ Feature: Reset passwords for one or more WordPress users.
...
+
+  @require-wp-4.3
+  Scenario: Reset the password of a WordPress user, and show only the new password
+    Given a WP installation
+
+    When I run `wp user get 1 --field=user_pass`
+    Then save STDOUT as {ORIGINAL_PASSWORD}
+
+    When I run `wp user reset-password 1 --skip-email --porcelain`
+    Then save STDOUT as {NEW_PASSWORD}
```
dsXLII, line 73: TODO: Learn enough behat to have this check the output (it should be a password, currently "exactly 12 characters, no whitespace, etc")

src/User_Command.php
```diff
@@ -1104,19 +1107,24 @@ public function import_csv( $args, $assoc_args ) {
...
 	public function reset_password( $args, $assoc_args ) {
 		$skip_email = Utils\get_flag_value( $assoc_args, 'skip-email' );
+		$show_new_pass = Utils\get_flag_value( $assoc_args, 'show-password' );
 		if ( $skip_email ) {
 			add_filter( 'send_password_change_email', '__return_false' );
 		}
 		$fetcher = new UserFetcher();
 		$users   = $fetcher->get_many( $args );
 		foreach ( $users as $user ) {
+			$new_pass = wp_generate_password();
```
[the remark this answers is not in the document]
dsXLII: Created a separate PR. :)

features/user-reset-password.feature
danielbachhuber: The check doesn't need to be super accurate. In other cases, we simply confirm that `STDOUT` is not empty.

Behat probably doesn't support checking "exactly 12 characters, no whitespace, etc". In order to perform such a check, you'd need to write a standalone PHP file to execute the check, pass the password to the PHP file, and then have the PHP file exit with 0 or 1 depending on whether it passed or failed. But, doing so is probably overkill for this particular test.

features/user-reset-password.feature
dsXLII: ... I have no idea what I was hung up on, disregard :)

dsmith4-godaddy:
I _think_ this is ready for a more thorough review. (Resolved the phpcs issues, and a merge conflict from the "change passwords to be 24 characters" PR.)

danielbachhuber (approved):
Great work on this, @dsXLII ! Sorry I missed your comment earlier.

I made some minor formatting tweaks with 6bb768194a02d3c144430012152d4f182586159d

PR #398, opened 2023-04-30
Add examples and update the documentation to the `wp post term`

Rahmon:
This PR tries to make more clear the usage of the command `wp post term` by adding more examples and updating the documentation.

### DIFF
--- a/src/WP_CLI/CommandWithTerms.php
+++ b/src/WP_CLI/CommandWithTerms.php
@@ -121,7 +121,7 @@ abstract class CommandWithTerms extends WP_CLI_Command {
 	 * : The name of the term's taxonomy.
 	 *
 	 * [<term>...]
-	 * : The slug of the term or terms (space-separated) to be removed from the object.
+	 * : The slug of the term or terms to be removed from the object.
 	 *
 	 * [--by=<field>]
 	 * : Explicitly handle the term value as a slug or id.
@@ -212,7 +212,7 @@ abstract class CommandWithTerms extends WP_CLI_Command {
 	 * : The name of the taxonomy type to be added.
 	 *
 	 * <term>...
-	 * : The slug of the term or terms (space-separated) to be added.
+	 * : The slug of the term or terms to be added.
 	 *
 	 * [--by=<field>]
 	 * : Explicitly handle the term value as a slug or id.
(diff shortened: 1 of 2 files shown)


src/WP_CLI/CommandWithTerms.php
```diff
@@ -121,11 +121,12 @@ public function list_( $args, $assoc_args ) {
 	 * : The name of the term's taxonomy.
 	 *
 	 * [<term>...]
-	 * : The name of the term or terms to be removed from the object.
+	 * : The slug of the term or terms (space-separated) to be removed from the object.
```
danielbachhuber, line 124:
```suggestion
	 * : The slug of the term or terms to be removed from the object.
```

"(space-separated)" isn't necessary here because it's the same behavior across all commands.

src/WP_CLI/CommandWithTerms.php
```diff
@@ -121,11 +121,12 @@ public function list_( $args, $assoc_args ) {
 	 * : The name of the term's taxonomy.
```
danielbachhuber, line 121: Should we update this one too? Are there others in the file?

src/WP_CLI/CommandWithTerms.php
Rahmon: Hi @danielbachhuber 

I've added `space-separated` through all commands so the user can have this information without having to visit another page. Do you think this is not necessary because of the examples added in the `wp post term` page?

src/WP_CLI/CommandWithTerms.php
Rahmon: I don't think so. By default, the slug is the taxonomy key but it can be different. I've updated the doc of the command `set` that was using `slug` instead of `name` to the `taxonomy` option.

src/WP_CLI/CommandWithTerms.php
danielbachhuber: > Do you think this is not necessary because of the examples added in the `wp post term` page?

Yes, and that it's standard behavior across all WP-CLI commands.

Rahmon: Got it. Removed.

danielbachhuber (approved):

PR #413, opened 2023-07-13
PHP 8.2 Deprecation: Fix creation of dynamic property.

Soean:
If I use WP CLI 2.8.1 with PHP 8.2.7 and call `wp site list`, I get a PHP deprecation message:

`Deprecated: Creation of dynamic property Site_Command::$fetcher is deprecated in phar:///usr/local/bin/wp/vendor/wp-cli/entity-command/src/Site_Command.php on line 37`

This PR adds properties to all classes with a fetcher.


swissspidy (approved):
Thank you!

Issue #268, opened 2019-09-30
wp post generate creates scheduled posts when the site timezone is ahead of UTC


johnbillion:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

When generating posts on a site where the `Timezone` setting is ahead of UTC, the timestamp appears correct but each post is scheduled by default, rather than published.

**Describe how other contributors can replicate this bug**

1. Set the site's timezone to something ahead of UTC, for example `wp option update timezone_string "Europe/Rome"` (or you can do this from the General Settings screen).
2. Run `wp post generate`
3. Visit the Posts screen in the admin area and note that all the generated posts are scheduled, despite having the correct post date.

The problem does not occur if the timezone is changed to `UTC+0`, or changed to a timezone behind UTC, for example `America/New_York`.

**Describe what you would expect as the correct outcome**

I would expect the posts to be published.

**Let us know what environment you are running this on**

```
OS:	Darwin 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64
Shell:	/bin/zsh
PHP binary:	/usr/local/Cellar/php/7.3.9_1/bin/php
PHP version:	7.3.9
php.ini used:	/usr/local/etc/php/7.3/php.ini
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/john
WP-CLI packages dir:	/Users/john/.wp-cli/packages/
WP-CLI global config:	/Users/john/.wp-cli/config.yml
WP-CLI project config:
WP-CLI version:	2.3.0
```

johnbillion:
I've tested this with WordPress 5.2.3 and with the latest trunk, and both exhibit the same result, but notably trunk marks the post as having missed its schedule.

danielbachhuber:
Looks like the problem still manifests in 6.2.2:

<img width="1446" alt="image" src="https://github.com/wp-cli/entity-command/assets/36432/b9212e25-80f4-456e-8c35-c970e2dbb7eb">

PR #418, opened 2023-08-24
Prevent unexpected missed schedules when generating new posts

rafaelzaleski:
This PR uses an empty string as the default value for the `--post_date` and `--post_date_gmt` parameters of `wp generate posts`. 
This will align with the behavior of `wp_insert_post` and use the current time for the generated posts when a date is not provided. If a date is provided, it's passed down to `wp_insert_post`.

Closes https://github.com/wp-cli/entity-command/issues/268
Related https://github.com/wp-cli/wp-cli/issues/5832

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -766,11 +766,13 @@ class Post_Command extends CommandWithDBObject {
 
 		// Handle edge cases for how WP handle dates on wp_insert_post for older versions. If one of the values is empty, it's assigned to the current time. We need to avoid sending an empty value.
 		if ( ! empty( $post_data['post_date'] ) && empty( $post_data['post_date_gmt'] ) ) {
-			$post_data['post_date_gmt'] = get_gmt_from_date( $post_data['post_date'] );
+			$only_date = strlen($post_data['post_date']) === 10;
+			$post_data['post_date_gmt'] = get_gmt_from_date( $post_data['post_date'], $only_date ? 'Y-m-d' : 'Y-m-d H:i:s' );
 		}
 
 		if ( ! empty( $post_data['post_date_gmt'] ) && empty( $post_data['post_date'] ) ) {
-			$post_data['post_date'] = get_date_from_gmt( $post_data['post_date_gmt'] );
+			$only_date = strlen($post_data['post_date_gmt']) === 10;
+			$post_data['post_date'] = get_date_from_gmt( $post_data['post_date_gmt'], $only_date ? 'Y-m-d' : 'Y-m-d H:i:s' );
 		}
 
 		if ( ! post_type_exists( $post_data['post_type'] ) ) {
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -765,11 +765,11 @@ class Post_Command extends CommandWithDBObject {
 		$post_data['post_date_gmt'] = $this->maybe_convert_hyphenated_date_format( $post_data['post_date_gmt'] );
 
 		// Handle edge cases for how WP handle dates on wp_insert_post for older versions. If one of the values is empty, it's assigned to the current time. We need to avoid sending an empty value.
-		if ( isset( $post_data['post_date'] ) && empty( $post_data['post_date_gmt'] ) ) {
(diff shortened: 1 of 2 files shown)


src/Post_Command.php
```diff
@@ -1006,4 +999,22 @@ public function exists( $args ) {
...
+	 * Convert a date-time string with a hyphen separator to a space separator.
+	 *
+	 * @param string $date_string The date-time string to convert.
+	 * @return string The converted date-time string.
+	 *
+	 * Example:
+	 * maybe_convert_hyphenated_date_format( "2018-07-05-17:17:17" );
+	 * Returns: "2018-07-05 17:17:17"
+	 */
+	private function maybe_convert_hyphenated_date_format( $date_string ) {
```
danielbachhuber, line 1013: Is this a new behavior? Can you share more detail about it?

danielbachhuber (changes requested):

src/Post_Command.php
rafaelzaleski: No, it's actually done to ensure compatibility with previous versions. The [docs](https://developer.wordpress.org/cli/commands/post/generate/) advises using a datetime string with a hyphen between the date and time. If we don't remove this hyphen before calling `wp_insert_post`, the timestamp will be saved as 0.

danielbachhuber: @rafaelzaleski Ok, great. Can you include some additional tests to capture this behavior then?

Also, it looks like your test introduces one break in a different test...

rafaelzaleski: > @rafaelzaleski Ok, great. Can you include some additional tests to capture this behavior then?

It's already [here](https://github.com/wp-cli/entity-command/blob/f6d53dfd0a8565b08f16a9633698b01170193cf6/features/post-generate.feature#L124) and [here](https://github.com/wp-cli/entity-command/blob/f6d53dfd0a8565b08f16a9633698b01170193cf6/features/post-generate.feature#L137).

> Also, it looks like your test introduces one break in a different test...

I'm not sure how to reproduce it locally. Do you have any ideas? 🤔

danielbachhuber: > I'm not sure how to reproduce it locally. Do you have any ideas?

I tried to get tests running against PHP 5.6 locally and failed 😕

First, I tried this [act](https://github.com/nektos/act) tool. It was able to run the unit tests but not the functional tests:

CleanShot 2023-09-08 at 07 29 06@2x

I think this is the specific failure:

CleanShot 2023-09-08 at 07 29 36@2x

Next, I tried to install PHP 5.6 directly on my machine:

```
brew install shivammathur/php/php@5.6
brew link --overwrite --force shivammathur/php/php@5.6
```

Something is funky with the tests, though:

CleanShot 2023-09-08 at 07 31 33@2x


I then manually tried to install WordPress 3.7, but received this MySQL error:

CleanShot 2023-09-08 at 07 34 56@2x

My guess is that something changed in the behavior of `wp_insert_post()` between WP 3.7 and today.

Based on the failed test, it looks like your code changes the result of `--post_date="2018-07-01"`` from `2018-07-01 00:00:00` to `1970-01-01 00:00:00`

rafaelzaleski: Thank you @danielbachhuber. I think I was able to fix the edge cases on WP 3.7 and I have requested a new review.

danielbachhuber: > I think I was able to fix the edge cases on WP 3.7

@rafaelzaleski Out of curiosity, how did you end up debugging/fixing this?

danielbachhuber (approved):
Nice work on this, @rafaelzaleski 👏

src/Post_Command.php
rafaelzaleski: > @rafaelzaleski Out of curiosity, how did you end up debugging/fixing this?

I encountered some challenges installing `3.7` locally too, mainly because of database issues.
 
So I went through the failed test data and reviewed the `wp_insert_post` source code from tag `3.7`. This helped me understand the nuances of the test data processing and the root cause of the failures. 
To ensure the solutions were working, I validated them using CI checks on my cloned repository.

PR #422, opened 2023-09-07
Fix typo for `$parent_instance` param in `RecursiveDataStructureTraverser` constructor

petitphp:

### DIFF
--- a/src/WP_CLI/Entity/RecursiveDataStructureTraverser.php
+++ b/src/WP_CLI/Entity/RecursiveDataStructureTraverser.php
@@ -24,9 +24,9 @@ class RecursiveDataStructureTraverser {
 	/**
 	 * RecursiveDataStructureTraverser constructor.
 	 *
-	 * @param mixed $data The data to read/manipulate by reference.
-	 * @param string|int $key The key/property the data belongs to.
-	 * @param static $parent_instance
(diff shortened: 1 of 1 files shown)


src/WP_CLI/Entity/RecursiveDataStructureTraverser.php
```diff
@@ -26,7 +26,7 @@ class RecursiveDataStructureTraverser {
 	 *
 	 * @param mixed $data The data to read/manipulate by reference.
 	 * @param string|int $key The key/property the data belongs to.
-	 * @param static $parent
+	 * @param static $parent_instance
```
schlessera, line 29: Let's fix the docblock fully while we're at it.
```suggestion
	 * @param static|null $parent_instance The parent instance of the traverser.
```
Also, could you adapt the alignment in the docblock? I cannot do that properly via a suggestion...

schlessera (changes requested):

src/WP_CLI/Entity/RecursiveDataStructureTraverser.php
danielbachhuber: Updated in 1c2e9fe0915053dc5ee861272b854adfca3f2dbc

danielbachhuber (approved):

petitphp:
Thanks @danielbachhuber !

PR #431, opened 2023-10-23
Improve SQLite compatibility

### LINKED ISSUE #430: SQLite Compatibility
When I manually run the tests locally with SQLite, the following scenarios are failing:

```
001 Scenario: Approving/unapproving comments # features/comment.feature:161
Then STDERR should contain:            # features/comment.feature:173
$ wp comment approve 2 --url=www.example.com
Success: Approved comment 2.

002 Scenario: Publishing a post and setting a date succeeds if the edit_date flag is passed. # features/post.feature:460
Then STDOUT should contain:                                                            # features/post.feature:473
$ wp post get 4 --field=post_date
2005-01-24T09:52:00.000Z

003 Scenario: Empty a site                                    # features/site-empty.feature:3
When I run `wp db query "SOURCE insert_link_data.sql;"` # features/site-empty.feature:15
$ wp db query "SOURCE insert_link_data.sql;"

ERROR 1146 (42S02) at line 1 in file: 'insert_link_data.sql': Table 'wp_cli_test.wp_links' doesn't exist

004 Scenario: Empty a site and its uploads directory # features/site-empty.feature:103
When I run `wp site empty --uploads --yes`     # features/site-empty.feature:114
$ wp site empty --uploads --yes
Success: The site at 'https://example.com' was emptied.

WordPress database error

[...]

Error occurred at line 3373 in Function <code>handle_error</code>. Error message was: SQLSTATE[HY000]: General error: 1 incomplete input.
Backtrace:
WP_SQLite_Translator->get_error_message()
#1 src/Site_Command.php(83): WP_SQLite_DB->query('TRUNCATE wp_pos...')
#2 src/Site_Command.php(263): Site_Command->empty_posts()
#3 [internal function]: Site_Command->empty_(Array, Array)

[...]

005 Scenario: Filter options by `--site_id`                                                                                      # features/site-option.feature:130
When I run `wp db query "INSERT INTO wp_sitemeta (site_id,meta_key,meta_value) VALUES (2,'wp_cli_test_option','foobar');"` # features/site-option.feature:133
$ wp db query "INSERT INTO wp_sitemeta (site_id,meta_key,meta_value) VALUES (2,'wp_cli_test_option','foobar');"

ERROR 1146 (42S02) at line 1: Table 'wp_cli_test.wp_sitemeta' doesn't exist

```

Let's figure out the best course of action here.

swissspidy:
As for `TRUNCATE TABLE`, this works better with the SQLite plugin right now. See https://github.com/WordPress/sqlite-database-integration/issues/51

Fixes #430
Related: https://github.com/wp-cli/wp-cli/issues/5859

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -476,6 +476,9 @@ Feature: Manage WordPress posts
       2005-01-24 09:52:00
       """
 
+  # Separate test because of a known bug in the SQLite plugin.
+  # See https://github.com/WordPress/sqlite-database-integration/issues/52.
+  # Once the bug is resolved, this separate test can be removed again.
   @require-sqlite
   Scenario: Publishing a post and setting a date succeeds if the edit_date flag is passed.
     Given a WP install
--- a/features/post.feature
+++ b/features/post.feature
@@ -457,6 +457,7 @@ Feature: Manage WordPress posts
       2005-01-24 09:52:00
       """
 
+  @require-mysql
   Scenario: Publishing a post and setting a date succeeds if the edit_date flag is passed.
     Given a WP install
 
@@ -473,4 +474,23 @@ Feature: Manage WordPress posts
     Then STDOUT should contain:
       """
       2005-01-24 09:52:00
+      """
+
+  @require-sqlite
+  Scenario: Publishing a post and setting a date succeeds if the edit_date flag is passed.
+    Given a WP install
+
+    When I run `wp post create --post_title='test' --porcelain`
+    Then save STDOUT as {POST_ID}
+
+    When I run `wp post update {POST_ID} --post_date='2005-01-24T09:52:00.000Z' --post_status='publish' --edit_date=1`
+    Then STDOUT should contain:
+      """
+      Success:
+      """
+
+    When I run `wp post get {POST_ID} --field=post_date`
+    Then STDOUT should contain:
+      """
+      2005-01-24T09:52:00.000Z
       """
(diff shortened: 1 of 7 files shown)


features/post.feature
```diff
@@ -473,4 +474,23 @@ Feature: Manage WordPress posts
...
+    When I run `wp post update {POST_ID} --post_date='2005-01-24T09:52:00.000Z' --post_status='publish' --edit_date=1`
+    Then STDOUT should contain:
+      """
+      Success:
+      """
+
+    When I run `wp post get {POST_ID} --field=post_date`
+    Then STDOUT should contain:
+      """
+      2005-01-24T09:52:00.000Z
```
schlessera, line 495: @swissspidy Seems to be a bug (in the SQLite plugin?) that the date is returned in a different format than with MySQL. What's the reasoning behind adding a separate test here, instead of fixing the behavior?

swissspidy: Yes it seems to be a bug, but in the SQLite plugin, so it should be fixed there. The thinking was that this test documents the existing behavior in the meantime, until that bug is fixed.

swissspidy: Bug filed here: https://github.com/WordPress/sqlite-database-integration/issues/52

swissspidy: We can also just remove this extra test for now, and simply skip the original one for SQLite until this bug is resolved. Don't have a strong opinion.

swissspidy: @schlessera What's your preference here?

schlessera: Lets keep the separate branch for now to unblock this but add a comment pointing to the bug report in the plugin.

swissspidy: Done!

schlessera (approved):

Issue #423, opened 2021-04-19
Support multiple roles in `user add-role` and `user remove-role`


mbucurcf:
### Feature Request
- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributions/contributing/).

#### Describe your use case and the problem you are facing

A new cli command for toggling user roles.

#### Describe the solution you'd like

It might be beneficial to add a command to toggle user roles. It could look like:
`wp user toggle-role user role1 role2 role3 ... rolen` 

- Notice that it could accept more than 1 role
- If a role is active, it will deactivated it
- If a role is inactive, it will activate it

PR #437, opened 2023-11-10
Support for adding and removing of multiple user roles

shawnhooper:
Fixes: https://github.com/wp-cli/entity-command/issues/423
Related: https://github.com/wp-cli/wp-cli/issues/5859

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -219,7 +219,8 @@ Feature: Manage WordPress users
 
     When I run `wp user add-role 1 editor`
     Then STDOUT should not be empty
-    And I run `wp user get 1 --field=roles`
+
+    When I run `wp user get 1 --field=roles`
     Then STDOUT should be:
       """
       administrator, editor
@@ -227,7 +228,8 @@ Feature: Manage WordPress users
 
     When I run `wp user add-role 1 editor contributor`
     Then STDOUT should not be empty
-    And I run `wp user get 1 --field=roles`
+
+    When I run `wp user get 1 --field=roles`
     Then STDOUT should be:
       """
       administrator, editor, contributor
@@ -235,7 +237,8 @@ Feature: Manage WordPress users
 
     When I run `wp user remove-role 1 editor contributor`
     Then STDOUT should not be empty
-    And I run `wp user get 1 --field=roles`
+
+    When I run `wp user get 1 --field=roles`
     Then STDOUT should be:
       """
       administrator
--- a/features/user.feature
+++ b/features/user.feature
@@ -218,7 +218,10 @@ Feature: Manage WordPress users
     And STDOUT should be empty
 
     When I run `wp user add-role 1 editor`
-    Then STDOUT should not be empty
+    Then STDOUT should be:
+      """
+      Success: Added 'editor' role for admin (1).
+      """
 
     When I run `wp user get 1 --field=roles`
     Then STDOUT should be:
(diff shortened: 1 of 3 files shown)


src/User_Command.php
```diff
@@ -692,26 +692,34 @@ public function set_role( $args, $assoc_args ) {
...
 	 * @subcommand add-role
 	 */
 	public function add_role( $args, $assoc_args ) {
 		$user = $this->fetcher->get_check( $args[0] );
 
-		$role = $args[1];
-
-		self::validate_role( $role );
+		$roles = $args;
+		array_shift( $roles );
```
danielbachhuber, line 713: Can we error early here if `$roles` is empty, and then make sure we have a test for the scenario too?

src/User_Command.php
```diff
@@ -722,27 +730,35 @@ public function add_role( $args, $assoc_args ) {
...
+	 *     Success: Removed 'editor' role for johndoe (12).
+	 *
 	 * @subcommand remove-role
 	 */
 	public function remove_role( $args, $assoc_args ) {
 		$user = $this->fetcher->get_check( $args[0] );
 
 		if ( isset( $args[1] ) ) {
-			$role = $args[1];
+			$roles = $args;
```
danielbachhuber, line 751: Same comment about erroring early if no role is provided.

danielbachhuber (changes requested):

src/User_Command.php
shawnhooper: @danielbachhuber No roles on the remove-role command is for multisite installs to remove the user from a blog.

src/User_Command.php
shawnhooper: @danielbachhuber Yep, I'll work on this now.

src/User_Command.php
danielbachhuber: @shawnhooper Ah, good point.

features/user.feature
```diff
@@ -209,6 +209,14 @@ Feature: Manage WordPress users
   Scenario: Managing user roles
     Given a WP install
 
+    When I run `wp user add-role 1`
```
shawnhooper, line 212: @danielbachhuber On my local, this test is failing, even though the feature works.  Any pointers as to what I did wrong here?

danielbachhuber:
```suggestion
    When I try `wp user add-role 1`
```

@shawnhooper It should be `When I try`, not `When I run`. 'try' is for error cases, 'run' is for success cases.

shawnhooper:
@danielbachhuber Ok, requested changes made and tests passing locally.  Ready for re-review.

src/User_Command.php
```diff
@@ -722,27 +734,35 @@ public function add_role( $args, $assoc_args ) {
...
-
-			$user->remove_role( $role );
+			foreach ( $roles as $role ) {
+				self::validate_role( $role );
+			}
 
-			WP_CLI::success( "Removed '{$role}' role for {$user->user_login} ({$user->ID})." );
+			foreach ( $roles as $role ) {
+				$user->remove_role( $role );
+				WP_CLI::success( "Removed '{$role}' role for {$user->user_login} ({$user->ID})." );
```
danielbachhuber, line 764: Rather than render `Success:` multiple times, can we consolidate into one `Success:`?

src/User_Command.php
```diff
@@ -692,26 +692,38 @@ public function set_role( $args, $assoc_args ) {
...
 
-		$user->add_role( $role );
+		foreach ( $roles as $role ) {
+			self::validate_role( $role );
+		}
 
-		WP_CLI::success( "Added '{$role}' role for {$user->user_login} ({$user->ID})." );
+		foreach ( $roles as $role ) {
+			$user->add_role( $role );
+			WP_CLI::success( "Added '{$role}' role for {$user->user_login} ({$user->ID})." );
```
danielbachhuber, line 725: Rather than render `Success:` multiple times, can we render it once and mention both roles?

danielbachhuber (changes requested):
Couple more tweaks, then should be good

danielbachhuber (approved):
🚢

PR #438, opened 2023-11-11
Add filter site__user_in on wp site list

marksabbath:
This is a HackDay PR 🎉 

This PR should cover https://github.com/wp-cli/ideas/issues/153

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -523,7 +523,7 @@ class Site_Command extends CommandWithDBObject {
 	 * [--site__in=<value>]
 	 * : Only list the sites with these blog_id values (comma-separated).
 	 *
-	 * [--user=<value>]
+	 * [--site_user=<value>]
 	 * : Only list the sites with this user.
 	 *
 	 * [--field=<field>]
@@ -611,8 +611,8 @@ class Site_Command extends CommandWithDBObject {
 			$where['site_id'] = $assoc_args['network'];
 		}
 
-		if ( isset( $assoc_args['user'] ) ) {
-			$user = ( new UserFetcher() )->get_check( $assoc_args['user'] );
+		if ( isset( $assoc_args['site_user'] ) ) {
+			$user = ( new UserFetcher() )->get_check( $assoc_args['site_user'] );
 
 			if ( $user ) {
 				$blogs = get_blogs_of_user( $user->ID );
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -612,7 +612,7 @@ class Site_Command extends CommandWithDBObject {
 		}
 
 		if ( isset( $assoc_args['user'] ) ) {
-			$user    = ( new UserFetcher() )->get_check( $assoc_args['user'] );
+			$user = ( new UserFetcher() )->get_check( $assoc_args['user'] );
 
 			if ( $user ) {
 				$blogs = get_blogs_of_user( $user->ID );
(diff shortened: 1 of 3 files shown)


src/Site_Command.php
```diff
@@ -520,7 +520,10 @@ private function get_network( $network_id ) {
 	 * 'url' isn't an available filter, as it comes from 'home' in wp_options.
 	 *
 	 * [--site__in=<value>]
-	 * : Only list the sites with these blog_id values (comma-separated).
+	 * : Only list 1 the sites with these blog_id values (comma-separated).
```
danielbachhuber, line 523:
```suggestion
	 * : Only list the sites with these blog_id values (comma-separated).
```

src/Site_Command.php
```diff
@@ -520,7 +520,10 @@ private function get_network( $network_id ) {
...
+	 *
+	 * [--site__user_in=<value>]
```
danielbachhuber, line 525: I don't love this argument name, but I don't have a great alternative to suggest yet. Maybe `--user__in=<value>`?

src/Site_Command.php
```diff
@@ -607,6 +610,24 @@ public function list_( $args, $assoc_args ) {
...
+			if ( $user ) {
+				$blogs = get_blogs_of_user( $user->id );
+
+				foreach ( $blogs as $blog ) {
+					$where['blog_id'][] = $blog->userblog_id;
+				}
+			}
+
+			if ( ! isset( $where['blog_id'] ) ) {
+				WP_CLI::error( 'Could not find a site with the user provided.' );
```
danielbachhuber, line 625: We should present an empty list here, instead of erroring.

src/Site_Command.php
```diff
@@ -607,6 +610,24 @@ public function list_( $args, $assoc_args ) {
 			$where['site_id'] = $assoc_args['network'];
 		}
 
+		if ( isset( $assoc_args['site__user_in'] ) ) {
+			$user = get_user_by( 'login', $assoc_args['site__user_in'] );
```
danielbachhuber, line 614: `_in` implies the argument supports a CSV of users. Should we accommodate that here?

Also, we should use `WP_CLI\Fetchers\User` instead of directly calling `get_user_by()`. It will handle the "invalid user" case for us.

danielbachhuber (changes requested):
Thanks for working on this, @marksabbath ! A few things to fix up.

src/Site_Command.php
marksabbath: I was on the fence with that as well. It _feels_ like `--user_in` would be a better name. I'll update the whole code to that.

src/Site_Command.php
marksabbath: Il'll take a look at `WP_CLI\Fetchers\User`.

danielbachhuber:
@marksabbath Are you still planning to finish this up?

marksabbath:
> @marksabbath Are you still planning to finish this up?

Yes! I'll make sure to spend some time this week.

src/Site_Command.php
marksabbath: I've updated the code to use `WP_CLI\Fetchers\User` and also updated the name of the flag to `--user`.

marksabbath: @danielbachhuber I needed to change the flag to `site_user` since it was not properly working with `user`. I guess it conflicts with some global flag or something.

danielbachhuber: > I needed to change the flag to `site_user` since it was not properly working with `user`. I guess it conflicts with some global flag or something.

@marksabbath Yep, it does. `site_user` is a fine alternative.

danielbachhuber (approved):
Looks great, @marksabbath ! Thanks for your work on this.

I added another test case with b0162070bbc5f6f899a38dc6e7d701c1238838f3

PR #442, opened 2023-11-14
Remove RecursiveDataStructureTraverser

schlessera:
This PR removes the `RecursiveDataStructureTraverser` class from the `wp-cli/entity-command` package and sets the package up to reuse it from the framework instead.

Note: Pushing first with the removal only to show how it fails before fixing the distribution aspect of it.


swissspidy (approved):

schlessera:
@danielbachhuber , @swissspidy some added context here to explain how this works:
The new `\WP_CLI\Traverser\RecursiveDataStructureTraverser` in the framework is not yet part of a tagged release, it is only included in the `main` branch so far.

However, given that:
* the stability requirements of this package have been set to `dev` => https://github.com/wp-cli/entity-command/blob/b5846bf9b14e7a0404ce9e50e6b64845bb02034b/composer.json#L199
* the `wp-cli/wp-cli` has a branch alias that is correctly set => https://github.com/wp-cli/wp-cli/blob/f37504278b39fc0265bf5a8936c345bc737ee0a1/composer.json#L40

... we can tell Composer to pull in version 2.10+ of the framework. It will accept `dev-main` to represent a non-stable version of 2.10, and the `dev` stability requirements will allow that version to be used for `wp-cli/entity-command`'s requirements => https://github.com/wp-cli/entity-command/pull/442/commits/a6d96ee1d1292c3a3c41366ae2e452e218033e21#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R15

This effectively means that the `wp-cli/entity-command` will pull in `dev-main` and assume it is v2.10+ (and all other packages have been set up to work correctly with this).

Then, once we will release a tagged version 2.10.0 for the `wp-cli/wp-cli` framework, the fact that the `wp-cli/entity-command` has `prefer-stable` set to `true` means that it will revert back to a stable, tagged version of 2.10.0 and all future versions as soon as possible => https://github.com/wp-cli/entity-command/blob/b5846bf9b14e7a0404ce9e50e6b64845bb02034b/composer.json#L200

/cc @thelovekesh

danielbachhuber:
@schlessera Thanks for explaining all of that!

PR #449, opened 2023-11-16
Add command to get the post ID by URL

JulianBustamante:
See #429 

- Adds url-to-id command to get the post ID from its URL.
- url_to_postid() doesn't check if the post exists when the URL contains the post id as a parameter.

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -264,13 +264,13 @@ Feature: Manage WordPress posts
     When I run `wp post url-to-id {POST_URL}`
     Then STDOUT should contain:
       """
-      {POST_ID}
+      1
       """
 
-    When I run `wp post url-to-id non-existent-url`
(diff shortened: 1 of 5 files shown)


features/post.feature
```diff
@@ -259,6 +259,13 @@ Feature: Manage WordPress posts
       """
       https://example.com/?p=1
       """
+    And save STDOUT as {POST_URL}
+
+    When I run `wp post url-to-id {POST_URL}`
+    Then STDOUT should contain:
+      """
+      {POST_ID}
+      """
```
danielbachhuber, line 268: Can we add a test for the error case too?

JulianBustamante: For sure!

danielbachhuber (approved):
Great work on this, @JulianBustamante !

Issue #322, opened 2021-06-21
Clarify that network meta and network options are the same thing


johnbillion:
## Feature Request

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

The `wp network meta` command is technically named correctly, but most developers know network meta as network options due to the `(add|get|update|delete)_network_option()` functions.

When you're trying to remember what command to use for network options WP-CLI isn't very helpful, especially for developers who don't know that network options are technically network meta.

**Describe the solution you'd like**

* Clarify the description of the `wp network meta` command so it mentions network options
* When a user types `wp network option`, show a suggestion for `wp network meta` (currently there's no suggestion)

shawnhooper:
@danielbachhuber I've tested out a patch that would work for suggestion `wp network meta` when you type `wp network option` but there's something about the approach I don't like.  

See:
https://github.com/wp-cli/wp-cli/commit/b6daeba7c0e9bab40fada1b9ea8017c5377e808e

It feels like putting this suggestion into the Runner feels too high-level for a very specific suggestion like this. 

But I don't see a good pattern for suggesting this command in the entity repo.

I also don't like putting this mapping in the Utils\get_suggestion() `$suggestion_map` array because I don't want to globally recommend replacing `option` with `meta`.  Only in the `network` command. 

Any suggested approaches before I turn this into a PR?

danielbachhuber:
@shawnhooper I think the Runner is the most appropriate place at this time. It's a one-off solution for a one-off idiosyncrasy. It will be safe to abstract further in the future if we need.

danielbachhuber:
🔨

PR #450, opened 2023-11-20
Feature test for `wp network option` suggestion

shawnhooper:
Relates to https://github.com/wp-cli/entity-command/issues/322

This PR adds a feature test for suggesting `wp network meta` if the user types `wp network option`.

NOTE:  This PR will fail until https://github.com/wp-cli/wp-cli/pull/5879 is merged. This suggestion is changed in the `wp-cli` repo - but the command to test it is in the `entity-command` repo.

### DIFF
--- a/features/network-meta.feature
+++ b/features/network-meta.feature
@@ -16,3 +16,14 @@ Feature: Manage network-wide custom fields.
       This is not a multisite install
       """
     And the return code should be 1
+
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@shawnhooper Let's add this feature test to https://github.com/wp-cli/wp-cli/pull/5879.

`wp-cli/wp-cli` requires `wp-cli/entity-command` as a dependency, so there shouldn't be any problem including it.

PR #458, opened 2024-01-17
Fix example for post create command

ernilambar:
Current:

```
# Create post and schedule for future
$ wp post create --post_type=page --post_title='A future post' --post_status=future --post_date='2020-12-01 07:00:00'
```

Example command does not match with its description.

Changed to:
```
# Create post and schedule for future
$ wp post create --post_type=post --post_title='A future post' --post_status=future --post_date='2030-12-01 07:00:00'
```


danielbachhuber (approved):
Thanks! 😅

PR #460, opened 2024-02-03
Support arrays/JSON in pluck and patch

### LINKED ISSUE #459: Option Pluck & Patch don't work with arrays or JSON
## Bug Report

The documentation for the "option pluck" and "option patch" commands suggest that they work with arrays and JSON. Specifically the reference to 'keys within the value'. However they do not.

Assuming my_option has a value of **a:1:{s:5:"myKey";s:6:"aValue";}**

The command `wp option pluck my_option myKey` returns nothing.

The command `wp option patch update my_option myKey newValue` returns 

> Error: No data exists for key "myKey"

Pluck:
- Create an option in WordPress with a serialized array
- Attempt to read a value from within that array e.g.;`wp option pluck my_option myKey`
- Notice the value is not returned

Patch
- Create an option in WordPress with a serialized array
- Attempt to write a new value for an entry within that array e.g.;`wp option patch update my_option myKey newValue`
- Notice that an error is returned and the value is not updated

It should be possible to read a value from an option stored as a serialized array or JSON.
It should be possible to write a new value into an option stored as a serialized array or JSON.

Windows 10, PHP8.2.15
```
OS:     Windows NT 10.0 build 19045 (Windows 10) AMD64
Shell:  C:\WINDOWS\system32\cmd.exe
PHP binary:     C:\PHP\php-8.2.15\php.exe
PHP version:    8.2.15
php.ini used:   C:\PHP\php-8.2.15\php.ini
MySQL binary:
MySQL version:
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       C:\Projects\CPRN\Website
WP-CLI packages dir:    C:\Users\rainc/.wp-cli/packages/
WP-CLI cache dir:       C:\Users\rainc/.wp-cli/cache
WP-CLI global config:   C:\Users\rainc/.wp-cli/config.yml
WP-CLI project config:
WP-CLI version: 2.9.0
```

Add detection for and handling of serialized values in options.
I have a pull request incoming.

**Provide additional context/Screenshots**

n/a

BlairCooper:
Add support to for serialized arrays and JSON to both pluck and patch. Add additional examples and a note about the stdin blocking issue with patch.

Fixes #459

### DIFF
--- a/features/option-pluck-patch.feature
+++ b/features/option-pluck-patch.feature
@@ -82,6 +82,83 @@ Feature: Option commands have pluck and patch.
       foo
       """
 
+  @pluck @pluck-array
+  Scenario: A value can be retrieved from an array by key.
+    Given a WP install
+    And I run `wp option update option_name 'a:1:{s:3:"key";s:5:"value";}'`
+
+    When I run `wp option pluck option_name key`
+    Then STDOUT should be:
+      """
+      value
+      """
+
+  @pluck @pluck-array
+  Scenario: A value can be retrieved from an array by index.
+    Given a WP install
+    And I run `wp option update option_name 'a:1:{i:0;s:5:"value";}'`
+
+    When I run `wp option pluck option_name 0`
+    Then STDOUT should be:
+      """
+      value
+      """
+
+  @pluck @pluck-array
+  Scenario: A value can be retrieved from an array within an array.
+    Given a WP install
+    And I run `wp option update option_name 'a:1:{s:3:"key";a:1:{i:0;s:10:"innerValue";}}'`
+
+    When I run `wp option pluck option_name key 0`
+    Then STDOUT should be:
+      """
+      innerValue
+      """
+
+  @pluck @pluck-array
+  Scenario: An array retrieved from an array by key.
+    Given a WP install
+    And I run `wp option update option_name 'a:1:{s:3:"key";a:1:{i:0;s:10:"innerValue";}}'`
+
+    When I run `wp option pluck option_name key`
+    Then STDOUT should be:
+      """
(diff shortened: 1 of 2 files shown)


danielbachhuber (changes requested):
Thanks for the pull request, @BlairCooper !

I don't think the `maybe_unserialize()` is necessary. `get_option()` is already unserializing for us.

I wasn't able to reproduce your original issue:

```
$ wp shell
wp> update_option( 'my_option', array( 'myKey' => 'myValue' ) );
=> bool(true)
$ wp option pluck my_option myKey
myValue
```

Notably, [in your original report](https://github.com/wp-cli/entity-command/issues/459), the string was incorrectly serialized:
> The documentation for the "option pluck" and "option patch" commands suggest that they work with arrays and JSON. Specifically the reference to 'keys within the value'. However they do not.
>
> Assuming my_option has a value of `a:1:{s:5:"myKey";s:6:"aValue"}`

The serialized value should be `a:1:{s:5:"myKey";s:7:"myValue";}`.

Additionally, adding support for JSON-encoded options (and meta, etc.) would be an enhancement. WordPress doesn't automatically work natively with JSON, so I'm not sure it makes sense to add that to WP-CLI.

Thoughts?

BlairCooper:
Concerning JSON, while WordPress doesn't support storing JSON natively having this functionality would be useful. WordPress does store the health-check-site-status-result (transient) option as a JSON serialized string.

I may revisit this pull request and break it into smaller pieces. Can you have a look at the feature tests that were added and let me know if they're valid scenario, or if there are others that should be added?

Issue #463, opened 2024-02-13
Position is not updated in wp menu item update command


ernilambar:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

Position is not updated in wp menu item update command.

**Describe how other contributors can replicate this bug**

`wp menu create "Sample Menu"`
`wp menu item add-custom sample-menu Alpha https://alpha.com`
`wp menu item add-custom sample-menu Beta https://beta.com`

`wp menu item list sample-menu --format=csv`

```
db_id,type,title,link,position
1272,custom,Alpha,https://alpha.com,1
1273,custom,Beta,https://beta.com,2
```

`wp menu item update 1273 --link=https://beta.net --position=1`

```
db_id,type,title,link,position
1272,custom,Alpha,https://alpha.com,1
1273,custom,Beta,https://beta.net,2
```

**Describe what you would expect as the correct outcome**

Position also should have been updated.

**Let us know what environment you are running this on**

```
OS:     Darwin 22.6.0 Darwin Kernel Version 22.6.0: Tue Nov  7 21:48:06 PST 2023; root:xnu-8796.141.3.702.9~2/RELEASE_X86_64 x86_64
Shell:  /bin/zsh
PHP binary:     /usr/local/Cellar/php@8.2/8.2.15/bin/php
PHP version:    8.2.15
php.ini used:   /usr/local/etc/php/8.2/php.ini
MySQL binary:   /usr/local/bin/mysql
MySQL version:  mysql  Ver 8.3.0 for macos13.6 on x86_64 (Homebrew)
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /Users/nilambarsharma/Sites/staging
WP-CLI packages dir:    /Users/nilambarsharma/.wp-cli/packages/
WP-CLI cache dir:       /Users/nilambarsharma/.wp-cli/cache
WP-CLI global config:   /Users/nilambarsharma/.wp-cli/config.yml
WP-CLI project config:  /Users/nilambarsharma/Sites/staging/wp-cli.local.yml
WP-CLI version: 2.10.0
```

[...]

PR #471, opened 2024-02-15
Fix position in menu item update

ernilambar:
Fixes https://github.com/wp-cli/entity-command/issues/463

 
- Fix position when updating menu item.

Eg: `wp menu item update 123 --position=2`

### DIFF
--- a/features/menu-item.feature
+++ b/features/menu-item.feature
@@ -195,3 +195,49 @@ Feature: Manage WordPress menu items
       | type   | title  | position | link               |
       | custom | First  | 1        | https://first.com  |
       | custom | Third  | 2        | https://third.com  |
+
+  Scenario: Menu order is updated after item update
+    When I run `wp menu create "Sidebar Menu"`
+    Then STDOUT should not be empty
+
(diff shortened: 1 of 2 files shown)


ernilambar:
@danielbachhuber Can you please review this approach of fixing menu item positions? I could not find a way without performing SQL query in loop. Any other alternative approach you have in mind? I will add Behat test after we finalize PHP code.

danielbachhuber:
@ernilambar How does WordPress core do it?

ernilambar:
@danielbachhuber Core does not have need to update separately. All data are fetched from $_POST and items are updated in loop. I also looked at the REST API. It has endpoint to update menu item separately but it also has issue regarding menu order and position. These menu and menu items seem like a different beast. 😰

danielbachhuber:
> I also looked at the REST API. It has endpoint to update menu item separately but it also has issue regarding menu order and position.

@ernilambar What is the issue? The same issue?

Feel free to start adding feature tests to this PR, if you'd like.

ernilambar:
This approach is not solving the exact issue. Need more research on this. Closing this for now.

Issue #472, opened 2024-02-16
Improve message in error trying to create user with existing email in multisite


ernilambar:
## Bug Report

- [✓] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [✓] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

When I try to create user in multisite with existing email address:

`wp user create admin1 admin1@gmail.com --role=administrator`

```
Error: <strong>Error:</strong> This email address is already registered. <a href="http://bahu.local/wp-login.php">Log in</a> with this address or choose another one.
```

**Describe what you would expect as the correct outcome**

```
Error: This email address is already registered.
```

**Let us know what environment you are running this on**

```
OS:     Darwin 22.6.0 Darwin Kernel Version 22.6.0: Tue Nov  7 21:48:06 PST 2023; root:xnu-8796.141.3.702.9~2/RELEASE_X86_64 x86_64
Shell:  /bin/zsh
PHP binary:     /usr/local/Cellar/php@8.2/8.2.15/bin/php
PHP version:    8.2.15
php.ini used:   /usr/local/etc/php/8.2/php.ini
MySQL binary:   /usr/local/bin/mysql
MySQL version:  mysql  Ver 8.3.0 for macos13.6 on x86_64 (Homebrew)
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /Users/nilambarsharma/Sites/bahu
WP-CLI packages dir:    /Users/nilambarsharma/.wp-cli/packages/
WP-CLI cache dir:       /Users/nilambarsharma/.wp-cli/cache
WP-CLI global config:
WP-CLI project config:  /Users/nilambarsharma/Sites/bahu/wp-cli.local.yml
WP-CLI version: 2.10.0
```

danielbachhuber:
Seems like a reasonable enhancement 👍

PR #473, opened 2024-02-21
Remove HTML tags from error message when creating user in multisite

ernilambar:
Fixes https://github.com/wp-cli/entity-command/issues/472

* Remove HTML tags from error message when creating user in multisite

Before, error message is like this:

```
Error: <strong>Error:</strong> This email address is already registered. <a href="http://bahu.local/wp-login.php">Log in</a> with this address or choose another one.
```

After: 
```
Error: This email address is already registered. Log in with this address or choose another one.
```


src/User_Command.php
```diff
@@ -429,7 +429,12 @@ public function create( $args, $assoc_args ) {
 		if ( is_multisite() ) {
 			$result = wpmu_validate_user_signup( $user->user_login, $user->user_email );
 			if ( is_wp_error( $result['errors'] ) && ! empty( $result['errors']->errors ) ) {
-				WP_CLI::error( $result['errors'] );
+				$message = $result['errors']->get_error_message();
+				if ( false !== stripos( $message, '</strong>' ) ) {
+					$exploded = explode( '</strong>', $message );
+					$message  = trim( wp_strip_all_tags( $exploded[1] ) );
+				}
```
swissspidy, line 436: This feels a bit too clever and does not explain what it does.

Why not simplify to:

```suggestion
				$message  = trim( wp_strip_all_tags( $message ) );
				$message  = str_replace( 'Error: ', '', $message );
```

swissspidy: Hmm then I'd rather lean towards not removing the double "Error: ". I don't think we do that anywhere else, so I don't see why we should do it here.

And doing `stripos` and `explode` is very fragile as the strings can change in WordPress. So I prefer keeping it simple.

But curious to hear what others think.

ernilambar: We have used similar in Plugin command. https://github.com/wp-cli/extension-command/blob/5619a55fd279ce016a299b898bbe8eedb2aecb53/src/Plugin_Command.php#L366 But it seems we have replaced english `Error:` here also.

ernilambar:
In WP 3.7 was `Sorry, that email address is already used!`. So, GitHub actions are not passing.

src/User_Command.php
ernilambar: May be we should use simple `$message  = trim( wp_strip_all_tags( $message ) ); $message  = str_replace( 'Error: ', '', $message );` and not worry about the string in Behat test.

swissspidy: Yeah WP-CLI is not usually translated

ernilambar:
Closing. We could explore more robust approach if needed.

Issue #474, opened 2024-02-23
Introduce ids as format for user application-password list


ernilambar:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

`wp user application-password list 1 --format=ids`

`--format=ids` could be helpful for piping commands. It will also bring parity with our listing commands.

**Describe the solution you'd like**

If output is like `8c4e8d37-7932-4ba4-8459-ead134a3db75 426eeda0-9c2a-4d29-aa7c-1e12c1c07629`, it can be helpful when using with xargs.

PR #475, opened 2024-02-23
Introduce `--format=ids` in user application password list

ernilambar:
Fixes https://github.com/wp-cli/entity-command/issues/474

### DIFF
--- a/src/User_Application_Password_Command.php
+++ b/src/User_Application_Password_Command.php
@@ -199,7 +199,7 @@ final class User_Application_Password_Command {
 		}
 
 		if ( ! empty( $assoc_args['format'] ) && 'ids' === $assoc_args['format'] ) {
-			$formatter = new Formatter( $assoc_args, $fields );
+			$formatter = new Formatter( $assoc_args );
 			$formatter->display_items( wp_list_pluck( $application_passwords, 'uuid' ) );
 		} else {
 			$formatter = new Formatter( $assoc_args, $fields );
--- a/src/User_Application_Password_Command.php
+++ b/src/User_Application_Password_Command.php
@@ -97,6 +97,7 @@ final class User_Application_Password_Command {
 	 *   - json
 	 *   - count
 	 *   - yaml
+	 *   - ids
 	 * ---
 	 *
 	 * [--orderby=<fields>]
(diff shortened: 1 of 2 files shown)


src/User_Application_Password_Command.php
```diff
@@ -197,8 +198,13 @@ static function ( $application_password ) use ( $field, $value ) {
...
 
-		$formatter = new Formatter( $assoc_args, $fields );
-		$formatter->display_items( $application_passwords );
+		if ( ! empty( $assoc_args['format'] ) && 'ids' === $assoc_args['format'] ) {
+			$formatter = new Formatter( $assoc_args, $fields );
+			$formatter->display_items( wp_list_pluck( $application_passwords, 'uuid' ) );
+		} else {
+			$formatter = new Formatter( $assoc_args, $fields );
+			$formatter->display_items( $application_passwords );
+		}
```
swissspidy, line 207: Nit: move `new Formatter()` up before the condition:

```suggestion
		$formatter = new Formatter( $assoc_args, $fields );

		if ( ! empty( $assoc_args['format'] ) && 'ids' === $assoc_args['format'] ) {
			$formatter->display_items( wp_list_pluck( $application_passwords, 'uuid' ) );
		} else {
			$formatter->display_items( $application_passwords );
		}
```

Aside: I noticed many commands just do `echo implode( ' ', wp_list_pluck( ... ) );` instead 🤔  Not sure why. Using the formatter is cleaner.

ernilambar: I agree. Those codes are from 7 or 8 years ago. Earlier such approach was used to print ids.

ernilambar: The change you mentioned did not work. So I have kept as it is.

swissspidy: How does moving `$formatter = new Formatter( $assoc_args, $fields );` up not work? 🤔 
It just removes a duplicated line by moving it out of the conditional block.

ernilambar: In Formatter value is passed by reference. So it messes up the actual value.

ernilambar:
@swissspidy In Formatter, `$assoc_args` is passed by reference. 

```
public function __construct( &$assoc_args, $fields = null, $prefix = false ) {
```
And it unsets the `format` key in the method. https://github.com/wp-cli/wp-cli/blob/main/php/WP_CLI/Formatter.php#L46

So, if we keep `$formatter = new Formatter( $assoc_args, $fields );` before the conditional, `if` condition will never be true because we now dont have `$assoc_args['format']`.

ernilambar:
cc @danielbachhuber

schlessera:
@ernilambar I suggest to go with the tweak that @swissspidy mentioned but also combine it with an early processing of the flag like so:
```
$format = Utils\get_flag_value( $assoc_args, 'format', 'table );
```
This will avoid the breakage with @swissspidy 's refactor, and it will also make the conditional(s) cleaner.

swissspidy (approved):

PR #476, opened 2024-02-23
Order in application password listing

ernilambar:


ernilambar:
After some digging I found out that, it is pretty rare edge case. In our test we have created two application passwords sequentially. Both got same value for `created` column. So, ordering of application passwords became fickle.

PR #478, opened 2024-02-23
Add `application-password` subcommands to readme

swissspidy:


composer.json
```diff
@@ -159,6 +159,13 @@
             "user",
             "user add-cap",
             "user add-role",
```
ernilambar, line 161: `site meta` and its subcommands are also missing.

Issue #461, opened 2024-02-10
Reassigning posts while deleting a user can result in lost posts if the target user does not exist


franz-josef-kaiser:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When deleting a user, associated posts get deleted as well. If one uses the `--reassign=<user-ID>`, the soon-to-be-deleted users posts get assigned to the user whose ID gets passed with the `--reassign` flag. The problem is that if the user ID points to a non-existing user, the posts are lost to a non-existing user.

**Describe the solution you'd like**

A possible fix would be to check if the user that should get the posts assigned does not exist, an exception gets thrown, the command does not proceed and the user gets informed that the target user does not exist. Maybe the cli could even tell to re-check the list of possible target users by using `wp users list`.

**Addendum**

The "fault" here is [not WP CLI (link to source)](https://github.com/wp-cli/entity-command/blob/2b5d5d54550edb7383ebaa77fb3a2d53871ba19a/src/User_Command.php#L302C17-L302C31), but [WordPress core command `wp_delete_user()`](https://developer.wordpress.org/reference/functions/wp_delete_user/) in its current 6.4 version:
```php
# source: wp_delete_user()
$wpdb->update( $wpdb->posts, array( 'post_author' => $reassign ), array( 'post_author' => $id ) );
```
WordPress does not perform a check as it assumes that the user was properly pre-selected. This is something that can, due to the absence of an GUI that might stop without a form value somewhere, hardly be assumed in an CLI environment.

A possible **Solution** might be to utilize the [`WP_User::get_data_by()`](https://developer.wordpress.org/reference/classes/wp_user/get_data_by/) method:
```patch
} else {
+ 	if ( ! WP_User::get_data_by( 'id', $reassign ) {
+ 		return ['error', $message = "No user with ID `{$user_id}` found.\nTry using `wp users list` to identify the correct user." ]
+ 	}
	$result  = wp_delete_user( $user_id, $reassign );
	$message = "Removed user {$user_id} from " . home_url() . '.';
}

danielbachhuber:
Thanks for the report, @franz-josef-kaiser 

Thinking out loud..

It's not so much that the data was lost though, right? The `post_author` simply ends up with a bad value, which you'd have to update with a SQL query.

[...]

PR #482, opened 2024-02-29
Validate reassign user in user delete command

ernilambar:
Fixes https://github.com/wp-cli/entity-command/issues/461

* This will perform sanity check for reassign value whether user is valid or not
* Asks for confirmation if not valid

### DIFF
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -280,8 +280,10 @@ class User_Command extends CommandWithDBObject {
 			WP_CLI::error( 'Reassigning content to a different user is not supported on multisite.' );
 		}
 
-		if ( ! $reassign ) {
-			WP_CLI::confirm( '--reassign parameter not passed. All associated posts will be deleted. Proceed?', $assoc_args );
+		$is_reassign_valid = ( false === get_userdata( $reassign ) ) ? false : true;
+
+		if ( ! $reassign || ! $is_reassign_valid ) {
+			WP_CLI::confirm( '--reassign parameter not passed or invalid. All associated posts will be deleted. Proceed?', $assoc_args );
 		}
 
 		$users = $this->fetcher->get_many( $args );
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -280,7 +280,7 @@ class User_Command extends CommandWithDBObject {
 			WP_CLI::error( 'Reassigning content to a different user is not supported on multisite.' );
 		}
 
-		$is_reassign_valid = ( false === get_userdata( $reassign ) ) ? false : true;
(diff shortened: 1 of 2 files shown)


ernilambar:
> Can we add tests for this?

@swissspidy I could not find a way to add feature test for this because behat test pauses due to `WP_CLI::confirm()` here.

src/User_Command.php
```diff
@@ -280,8 +280,10 @@ public function delete( $args, $assoc_args ) {
 			WP_CLI::error( 'Reassigning content to a different user is not supported on multisite.' );
 		}
 
-		if ( ! $reassign ) {
-			WP_CLI::confirm( '--reassign parameter not passed. All associated posts will be deleted. Proceed?', $assoc_args );
+		$is_reassign_valid = ( false === get_userdata( $reassign ) ) ? false : true;
+
+		if ( ! $reassign || ! $is_reassign_valid ) {
+			WP_CLI::confirm( '--reassign parameter not passed or invalid. All associated posts will be deleted. Proceed?', $assoc_args );
```
danielbachhuber, line 286: I think we should actually have two separate confirmation prompts, so we can use more precise messages:

```php
if ( ! $is_reassign_valid ) {
	WP_CLI::confirm( '--reassign parameter is invalid. All associated posts will be deleted. Proceed?', $assoc_args );
}

if ( ! $reassign ) {
	WP_CLI::confirm( '--reassign parameter not passed. All associated posts will be deleted. Proceed?', $assoc_args );
}
```

Here's an example where session input is passed to the command:

https://github.com/wp-cli/entity-command/blob/fd187722baecc28b955b0a42d0b780042ca6e896/features/post-meta-clean-duplicates.feature#L33

You can follow a similar approach for writing a test here.

danielbachhuber (changes requested):

src/User_Command.php
ernilambar: PR has been updated.

src/User_Command.php
```diff
@@ -280,8 +280,12 @@ public function delete( $args, $assoc_args ) {
 			WP_CLI::error( 'Reassigning content to a different user is not supported on multisite.' );
 		}
 
+		$is_reassign_valid = ( false === get_userdata( $reassign ) ) ? false : true;
```
danielbachhuber, line 283:
```suggestion
		$is_reassign_valid = ( $reassign && false === get_userdata( $reassign ) ) ? false : true;
```

Let's avoid the DB hit unless `$reassign` is provided.

danielbachhuber (approved):

danielbachhuber:
Great work on this PR, @ernilambar !

Issue #479, opened 2024-02-26
Support username and email also as ID in wp user spam


ernilambar:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

In our all `wp user` commands, we accept username, user email or user ID as ID except `wp user spam` and `wp user unspam`. We could improve that to make it uniform. 

**Describe the solution you'd like**

`wp user spam username` should also work.

PR #485, opened 2024-03-05
Accept user login and email in spam and unspam commands

ernilambar:
In fact our code already supports user logins. So I have updated only docs to reflect this.

Fixes https://github.com/wp-cli/entity-command/issues/479


danielbachhuber (approved):
Thanks!

PR #486, opened 2024-03-05
Add user exists command

### LINKED ISSUE #484: Add a `wp user exists` subcommand
## Feature Request

Using `wp user get` writes to stderr when the user doesn't exist, which makes it porcelain, since it needs to be parsed and processed to be able to distinguish between user doesn't exist and unrelated errors.

There should be a command like `wp user exists` to check whether a user exists or not, returning 0 on success, 1 if it doesn't exist and not write anything to stderr in that case.
Like [wp plugin is-active](https://developer.wordpress.org/cli/commands/plugin/is-active/) which does this for plugins.

shawnhooper:
Adds a user  `wp user exists` command that works the same way `wp post exists` does.

Fixes #484

### DIFF
--- a/README.md
+++ b/README.md
@@ -4769,6 +4769,36 @@ make sure to reassign their posts prior to deleting the user.
 
 
 
+### wp user exists
+
+Verifies whether a user exists.
+
+~~~
+wp user exists <id>
+~~~
+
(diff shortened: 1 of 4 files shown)


danielbachhuber (approved):
Thanks!

Issue #274, opened 2019-11-06
wp user create ignores --nickname


Hlsgs:
`wp user create` ignores `--nickname` and always uses  the username instead

johnbillion:
It looks like this affects other fields too, such as `--user_nicename` and `--user_url`. See #301 .

schlessera:
Weirdly enough, the `nickname` is only available when you retrieve users via `get_users()`, and not when you retrieve them via `get_user_by()`... See https://github.com/wp-cli/wp-cli/pull/5600

Hlsgs:
Confirming still present in wp cli v2.6.0

schlessera:
@Hlsgs Yes, indeed. I did some work in solving this, but it turns around it is mostly a bug in `get_users()`. I then tried to use `WP_User_Query` across the board for all WP-CLI user queries, but that just turned up with different bugs of its own.
This needs more careful investigation in how best to solve.

Hlsgs:
> @Hlsgs Yes, indeed. I did some work in solving this, but it turns around it is mostly a bug in `get_users()`. I then tried to use `WP_User_Query` across the board for all WP-CLI user queries, but that just turned up with different bugs of its own. This needs more careful investigation in how best to solve.

Got it! Thank you for the lowdown.

PR #494, opened 2024-04-19
Use TwentyTwelve theme in menu location test

ernilambar:
Related https://github.com/wp-cli/wp-cli-tests/issues/200

* Use TwentyTwelve theme in menu location test

PR #495, opened 2024-04-23
Support nickname creating user

ernilambar:
Fixes https://github.com/wp-cli/entity-command/issues/274

* Support nickname creating user
* Add tests for user creation with nickname


swissspidy:
Mind rebasing?

swissspidy:
The `WP trunk` tests will still fail, but at least the `WP latest` tests will pass :-)

ernilambar:
Hmm. It seems for Latest also https://github.com/wp-cli/entity-command/pull/494 needs to be merged first.

PR #496, opened 2024-04-25
Support new upstream autoload options

### LINKED ISSUE #491: Update `wp option` for new changes in trunk / WordPress 6.6
## Feature Request

https://core.trac.wordpress.org/changeset/57920 got committed in trunk yesterday.

This caused some wp option tests to fail on trunk, see https://github.com/wp-cli/entity-command/actions/runs/8563704484/job/23469200909

WP-CLI should support the new logic and tests should pass again.

mrsdizzie:
Fixes #491

See: https://core.trac.wordpress.org/changeset/57920

This should get all of the tests working again

### DIFF
--- a/features/option.feature
+++ b/features/option.feature
@@ -140,6 +140,7 @@ Feature: Manage WordPress options
       """
 
   @require-wp-4.2
+  @less-than-wp-6.6
   Scenario: Update autoload value for custom option
     Given a WP install
     And I run `wp option add hello world --autoload=no`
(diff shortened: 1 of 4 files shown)


features/option.feature
```diff
@@ -248,11 +272,11 @@ Feature: Manage WordPress options
       """
     And the return code should be 0
 
-    When I try `wp option list --search='auto_opt' --autoload=no`
+    When I try `wp option list --search='auto_opt' --autoload=nope`
     Then STDOUT should be empty
-    And STDERR should be:
+    And STDERR should contain:
```
mrsdizzie, line 277: I changed this to contain because the error message has been updated (to include yes and no as valid options) and it didn't seem worth an entire extra Scenario for wp 6.6

I can add one if wanted

swissspidy: Sounds fine to me 👍

swissspidy (approved):

PR #498, opened 2024-04-26
Add site generator command

### LINKED ISSUE #231: Add support for generating sites
When creating a multisite instance, it would be helpful to have a `wp site generate` command to easily create a large multisite environment for testing.

i-am-chitti:
## Description

Add site generator command
```
wp site generate
```

### Options
- slug
- count
- email
- network_id
- private
- progress

### Issues
Fixes #231 
Related https://github.com/wp-cli/wp-cli/issues/5935

### DIFF
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -552,8 +552,7 @@ class Site_Command extends CommandWithDBObject {
 		$is_subdomain_install = is_subdomain_install();
 		// If not a subdomain install, make sure the domain isn't a reserved word
 		if ( ! $is_subdomain_install ) {
-			// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound -- Calling WordPress native hook.
-			$subdirectory_reserved_names = apply_filters( 'subdirectory_reserved_names', [ 'page', 'comments', 'blog', 'files', 'feed' ] );
+			$subdirectory_reserved_names = get_subdirectory_reserved_names();
 			if ( in_array( $base, $subdirectory_reserved_names, true ) ) {
 				WP_CLI::error( 'The following words are reserved and cannot be used as blog names: ' . implode( ', ', $subdirectory_reserved_names ) );
 			}
--- a/src/Site_Command.php
+++ b/src/Site_Command.php
@@ -488,6 +488,148 @@ class Site_Command extends CommandWithDBObject {
 	}
 
 	/**
+	 * Generate some sites.
+	 *
+	 * Creates a specified number of new sites.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--count=<number>]
+	 * : How many sites to generates?
+	 * ---
+	 * default: 100
+	 * ---
+	 *
+	 * [--email=<email>]
+	 * : Email for admin user. User will be created if none exists. Assignment to super admin if not included.
+	 *
+	 * [--network_id=<network-id>]
+	 * : Network to associate new site with. Defaults to current network (typically 1).
+	 *
+	 * [--private]
+	 * : If set, the new site will be non-public (not indexed)
+	 *
+	 * [--format=<format>]
+	 * : Render output in a particular format.
+	 * ---
+	 * default: progress
+	 * options:
+	 *  - progress
+	 *  - ids
+	 * ---
+	 *
+	 * ## EXAMPLES
+	 *
+	 *    # Generate 10 sites.
+	 *    $ wp site generate --count=10
+	 *    Generating sites  100% [================================================] 0:01 / 0:04
+	 */
+	public function generate( $args, $assoc_args ) {
+		if ( ! is_multisite() ) {
+			WP_CLI::error( 'This is not a multisite installation.' );
+		}
+
+		global $wpdb, $current_site;
+
+		$defaults = [
+			'count'      => 100,
+			'email'      => '',
+			'network_id' => 1,
+		];
+
+		$assoc_args = array_merge( $defaults, $assoc_args );
+
+		// Network.
+		if ( ! empty( $assoc_args['network_id'] ) ) {
+			$network = $this->get_network( $assoc_args['network_id'] );
+			if ( false === $network ) {
+				WP_CLI::error( "Network with id {$assoc_args['network_id']} does not exist." );
+			}
+		} else {
+			$network = $current_site;
+		}
+
+		// Public.
+		$public = ! Utils\get_flag_value( $assoc_args, 'private' );
+
+		// Limit.
+		$limit = $assoc_args['count'];
+
+		// Email.
+		$email = sanitize_email( $assoc_args['email'] );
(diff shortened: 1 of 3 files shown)


swissspidy:
Neat!

With this being a new command, let's add it to this list here:

https://github.com/wp-cli/entity-command/blob/8b19b7dacca26dea8f675e394230c63ebee1277d/composer.json#L125-L130

That will make sure the readme can be updated properly. The readme will be updated automatically via a cron workflow, but you can also update it manually in this PR using `wp scaffold package-readme` if you'd like to see the changes.

src/Site_Command.php
```diff
@@ -487,6 +487,166 @@ public function create( $args, $assoc_args ) {
...
+		$base = $assoc_args['slug'];
+		if ( preg_match( '|^([a-zA-Z0-9-])+$|', $base ) ) {
+			$base = strtolower( $base );
+		}
+
+		$is_subdomain_install = is_subdomain_install();
+		// If not a subdomain install, make sure the domain isn't a reserved word
+		if ( ! $is_subdomain_install ) {
+			// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound -- Calling WordPress native hook.
+			$subdirectory_reserved_names = apply_filters( 'subdirectory_reserved_names', [ 'page', 'comments', 'blog', 'files', 'feed' ] );
```
swissspidy, line 556: Can we just use the `get_subdirectory_reserved_names()` function here?

src/Site_Command.php
```diff
@@ -487,6 +487,166 @@ public function create( $args, $assoc_args ) {
...
+				}
+			} else {
+				WP_CLI::error( $id->get_error_message() );
+			}
+
+			if ( 'progress' === $format ) {
+				$notify->tick();
+			} else {
+				echo $id;
+				if ( $index < $limit -1 ) {
```
swissspidy, line 639:
```suggestion
				if ( $index < $limit - 1 ) {
```

src/Site_Command.php
```diff
@@ -487,6 +487,166 @@ public function create( $args, $assoc_args ) {
...
+		$format = Utils\get_flag_value( $assoc_args, 'format', 'progress' );
+
+		$notify = false;
+		if ( 'progress' === $format ) {
+			$notify = Utils\make_progress_bar( 'Generating sites', $limit );
+		}
+
+		for ( $index = 1; $index <= $limit; $index++ ) {
+			$current_base  = $base . $index;
+			$title = ucfirst( $base ) . ' ' . $index;
```
swissspidy, line 613:
```suggestion
			$current_base = $base . $index;
			$title        = ucfirst( $base ) . ' ' . $index;
```

i-am-chitti:
@swissspidy Addressed the feedback. Please have a re-review. 
Also, the Behat tests are failing for the newly added site-generate.feature . It's passing on my local - 
image

In CI -
image

swissspidy:
I don't see that test failing on CI, I only see this error:

```
PHP Fatal error:  Call to undefined function get_subdirectory_reserved_names() in src/Site_Command.php on line 555
```

That's because of my suggestion to use `get_subdirectory_reserved_names()`. Turns out this was only added in WP 4.4, but we still run tests against 3.7+.

To address this, we can add a new private method to the class like this:

```php
/**
 * Retrieves a list of reserved site on a sub-directory Multisite installation.
 *
 * Works on older WordPress versions where get_subdirectory_reserved_names() does not exist.
 *
 * @return string[] Array of reserved names.
 */
private function get_subdirectory_reserved_names() {
	if ( function_exists( 'get_subdirectory_reserved_names' ) ) {
		return get_subdirectory_reserved_names();
	}

	$names = array(
		'page',
		'comments',
		'blog',
		'files',
		'feed',
		'wp-admin',
		'wp-content',
		'wp-includes',
		'wp-json',
		'embed',
	);

	// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound -- Calling WordPress native hook.
	return apply_filters( 'subdirectory_reserved_names', $names );
}
```

swissspidy:
Ah, now I see the failing test correctly, again only on older WP versions. https://github.com/wp-cli/entity-command/actions/runs/8852432605/job/24311451054?pr=498#step:11:86

Unfortunately I don't have a local 3.7 test site (or an old PHP version) right now to manually test this.

I wonder if it has to do with the URL scheme (https vs http). Some tests do

```
    And I run `wp site list --site__in={SITE_ID} --field=url | sed -e's,^\(.*\)://.*,\1,g'`
    And save STDOUT as {SCHEME}
```

And then use e.g. `{SCHEME}://example.com/site1/` in the test.

If it's not that, then maybe someone else has an idea.

---

Aside: any idea why this one other unrelated test could be failing only on this PR?

https://github.com/wp-cli/entity-command/actions/runs/8852432605/job/24311451385?pr=498#step:11:84

ernilambar:
@swissspidy 
https://github.com/wp-cli/entity-command/actions/runs/8852432605/job/24311451385?pr=498#step:11:84

Regarding this error, I have also seen this while working in the PR which was merged yesterday.

We have:

```
When I run `wp user application-password create 1 myapp1`
Then STDOUT should not be empty

When I run `wp user application-password create 1 myapp2 --app-id=42`
Then STDOUT should not be empty
```

We are creating two application passwords here. If the timestamp is exactly same for both entries then order of entries is not predictable for following command. And then this error appears.

```
wp user application-password list 1 --format=ids --orderby=created --order=asc
```

Should we use `sleep` command before creating second application password in the test so that timestamp for two entries here will never be same?

swissspidy:
I see.

We should avoid things like`sleep`.

There must be other ways to make the test more robust. For example:

a) remove the `orderby=created` test case as it doesn't seem to provide too much value
b) rewrite the test to do something more like this:

```
When I run `wp user application-password list admin --orderby=created --order=asc --format=csv`
Then STDOUT should contain:
	"""
	{UUID1},,myapp1,{HASH1},{CREATED1},,,,
	"""
And STDOUT should contain:
	"""
	{UUID1},42,myapp12{HASH2},{CREATED2},,,,
	"""
```

Wanna open a PR?

ernilambar:
@swissspidy https://github.com/wp-cli/entity-command/pull/499

swissspidy:
@wp-cli/committers Does anyone have thoughts here on how to debug the WP 3.7 failure? I don't have a working local setup to test it myself.

ernilambar:
Looks like we faced same issue and fixed using `{SCHEME}` few years ago. https://github.com/wp-cli/entity-command/commit/e0a00c555f6a4b1ab467460b2c306aaa5b2157b7 May be we should use same approach here also.

i-am-chitti:
Thanks @ernilambar. This is fixing the failing tests on local. Hoping to get the same results in CI.

swissspidy (approved):

ernilambar:
@swissspidy Any idea why regenerate-readme was not triggered after this PR?

PR #502, opened 2024-05-16
Fix menu ordering on item update

nateinaction:
Fixes https://github.com/wp-cli/entity-command/issues/463

This PR reorders menu items when using `wp menu item update` command with the `--position` flag

Previous behavior is described in #463

New behavior:
```sh
$ wp menu item list sample-menu --format=csv
db_id,type,title,link,position
4,custom,Alpha,https://alpha.com,1
5,custom,Beta,https://beta.com,2

$ wp menu item update 5 --position=1
Success: Menu item updated.

$ wp menu item list sample-menu --format=csv
db_id,type,title,link,position
5,custom,Beta,https://beta.com,1
4,custom,Alpha,https://alpha.com,2

$ wp menu item update 5 --position=2        
Success: Menu item updated.

$ wp menu item list sample-menu --format=csv
db_id,type,title,link,position
4,custom,Alpha,https://alpha.com,1
5,custom,Beta,https://beta.com,2
```

While this fix works, it utilizes `reorder_menu_items()` which I believe is potentially error prone in situations where there are gaps in menu item ordering. Menu ordering gaps can be introduced using the same `wp menu item update` command and can cause abnormal behavior when reordering position. I'm curious about the maintainer's interest in enforcing menu item ordering without allowing gaps?

Another potential bug could be the `position` shown when listing the menu does not match the `menu_order` column in the `posts` table. By not displaying the exact `menu_order` value we can cause [confusion like I had](https://github.com/wp-cli/entity-command/issues/463#issuecomment-2116104111) when investigating this issue.

### DIFF
--- a/features/menu-item.feature
+++ b/features/menu-item.feature
@@ -195,3 +195,49 @@ Feature: Manage WordPress menu items
       | type   | title  | position | link               |
       | custom | First  | 1        | https://first.com  |
       | custom | Third  | 2        | https://third.com  |
+
+  Scenario: Menu order is updated after item update
+    When I run `wp menu create "Sidebar Menu"`
+    Then STDOUT should not be empty
+
+    When I run `wp menu item add-custom sidebar-menu First https://first.com --porcelain`
+    Then save STDOUT as {ITEM_ID_1}
+
+    When I run `wp menu item add-custom sidebar-menu Second https://second.com --porcelain`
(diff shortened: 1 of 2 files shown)


nateinaction:
This is close but we need slightly more complex logic for reordering. Still seeing odd ordering issues on top of problems when gaps exist in menu item order.

Issue #509, opened 2024-08-16
`wp option` - autoload filter doesn't display new autoload values


jhagrid77:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

The `option list` command with the `--autoload` option does not search for the [new `on`/`off` values in WordPress Core 6.6](https://core.trac.wordpress.org/changeset/57920), resulting in no return values.

**Describe how other contributors can replicate this bug**

- perform a fresh installation of WordPress Core 6.6.1
- download the latest WP-CLI version, 2.11
- run `[path to wp-cli.phar] option list --autoload=on`
- run `[path to wp-cli.phar] option list --autoload=yes`
- run `[path to wp-cli.phar] option list --autoload=off`
- run `[path to wp-cli.phar] option list --autoload=no`

**Describe what you would expect as the correct outcome**

I would expect that either:
- searching for on/yes (and off/no) would search for both autoload values
- the new autoload values (on/off) would have their own search values

**Let us know what environment you are running this on**

```
$ ./wp-cli.phar cli info
OS:     Linux 5.4.245-200.el7.x86_64 #1 SMP Thu Jun 8 15:58:31 EDT 2023 x86_64
Shell:  /bin/bash
PHP binary:     /opt/remi/php81/root/usr/bin/php
PHP version:    8.1.28
php.ini used:   /etc/opt/remi/php81/php.ini
MySQL binary:   /bin/mysql
MySQL version:  mysql  Ver 15.1 Distrib 10.5.23-MariaDB, for Linux (x86_64) using readline 5.1
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /chroot/home/ab59ee70/5cb513e4ba.nxcli.io/html
WP-CLI packages dir:
WP-CLI cache dir:       /home/ab59ee70/.wp-cli/cache
WP-CLI global config:   /home/ab59ee70/.wp-cli/config.yml
WP-CLI project config:
WP-CLI version: 2.11.0

$ ./wp-cli.phar core version
6.6.1
```

**Provide a possible solution**

[...]

PR #515, opened 2024-10-22
`wp option` - new autoload values in autoload filter

imrraaj:
- Replaces the old values of autoload filter to new ones when passing via flag in the `wp option list` command
- Checks for the autoload values of `on` and `off` instead of old values `yes` and `no`.
- Fixes: #509

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "wp-cli/wp-cli": "^2.11"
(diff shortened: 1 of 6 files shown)


swissspidy:
Hi there and thanks for your contribution!

As per the reported issue, we should search for both values, old and new, not just the new ones.

Also, would it be possible to add tests for this change?

mrsdizzie (approved):

PR #518, opened 2025-01-22
Implement network-specific metadata handling in commands

### LINKED ISSUE #504: Network meta command accessing the incorrect caches
## Bug Report

- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

Originally reported https://core.trac.wordpress.org/ticket/61467.

> It appears that there are multiple ways to work with wp_sitemeta and each use different cache keys and groups, resulting in inconsistent behavior and out-of-sync cached values.
One method is using update_site_option()/update_network_option() which uses "$network_id:$option" for the cache key, and site-options for the group.
The other method is update_metadata( 'site', ... ) which uses the object ID (here the network ID) for the cache key, and site_meta for the cache group.
Normally, in the code, we use update_site_option()/update_network_option(), however, WP CLI's wp network meta ... uses the *_metadata() functions, and when using both we end up with out of sync cached values.

Steps to reproduce the issue via WP CLI:

```
wp eval 'update_network_option( 1, "mykey", "123" );'
wp eval 'echo get_network_option( 1, "mykey" );' #Displays 123

wp network meta update 1 mykey 456
wp network meta get 1 mykey #Displays 456

wp eval 'echo get_network_option( 1, "mykey" );' #Still displays 123, but 456 is the expected value
```

```js
// You can also use code snippets if needed.
```

That network option caches are used and not meta data caches. This means for those using object caches that presist, that the correct cache is updated.

```
(Paste the output of "wp cli info" into this box)
```

Use network option function / api instead of metadata apis for CRUD. 

```
class Network_Meta_Command extends CommandWithMeta {
	protected $meta_type = 'site';

	protected function get_metadata( $object_id, $meta_key = '', $single = false ) {
		return get_network_option( $object_id, $meta_key );
	}

        protected function delete_metadata( $object_id, $meta_key, $meta_value = '' ) {
		return delete_network_option( $object_id, $meta_key );
	}

        protected function add_metadata( $object_id, $meta_key, $meta_value, $unique = false ) {
		return add_network_option( $object_id, $meta_key, $meta_value );
	}

	protected function update_metadata( $object_id, $meta_key, $meta_value, $prev_value = '' ) {
		return update_network_option( $object_id, $meta_key, $meta_value );
	}
}
```

This would require WordPress 4.4 as a min supported version.

spacedmonkey:
Added overrides for add, update, get, and delete metadata methods to utilize network-specific options when available. This ensures compatibility and functionality for multisite network scenarios. Fallbacks to standard metadata functions are maintained for non-network environments.

Add function exists, to ensure compatablity with 4.2 below. 

Fixes #504

### DIFF
--- a/features/network-meta.feature
+++ b/features/network-meta.feature
@@ -17,7 +17,6 @@ Feature: Manage network-wide custom fields.
       """
     And the return code should be 1
 
-  # TODO: FIXME
   Scenario: Network meta is actually network options
     Given a WP multisite install
 
@@ -44,4 +43,42 @@ Feature: Manage network-wide custom fields.
     Then STDOUT should be:
       """
       456
+      """
+
+  @require-object-cache
+  Scenario: Object cache correctly handles network meta updates
+    Given a WP multisite install
+
+    When I run `wp eval 'update_network_option( 1, "objkey", "123" );'`
+
+    When I run `wp network meta get 1 objkey`
+    Then STDOUT should be:
+      """
+      123
+      """
+
+    When I run `wp eval 'update_network_option( 1, "objkey", "456" );'`
+
+    When I run `wp network meta get 1 objkey`
+    Then STDOUT should be:
+      """
+      456
+      """
+
+    When I run `wp network meta update 1 objkey 789`
+    Then STDOUT should be:
+      """
+      Success: Updated custom field 'objkey'.
+      """
+
+    When I run `wp network meta get 1 objkey`
+    Then STDOUT should be:
+      """
+      789
+      """
+
+    When I run `wp eval 'echo get_network_option( 1, "objkey" );'`
+    Then STDOUT should be:
+      """
+      789
       """
--- a/features/network-meta.feature
+++ b/features/network-meta.feature
@@ -16,3 +16,32 @@ Feature: Manage network-wide custom fields.
       This is not a multisite install
       """
     And the return code should be 1
+
+  # TODO: FIXME
+  Scenario: Network meta is actually network options
+    Given a WP multisite install
+
+    When I run `wp eval 'update_network_option( 1, "mykey", "123" );'`
+    And I run `wp eval 'echo get_network_option( 1, "mykey" );'`
+    Then STDOUT should be:
+      """
+      123
+      """
+
+    When I run `wp network meta update 1 mykey 456`
+    Then STDOUT should be:
+      """
+      Success: Updated custom field 'mykey'.
+      """
+
+    When I run `wp network meta get 1 mykey`
+    Then STDOUT should be:
+      """
+      456
+      """
+
+    When I run `wp eval 'echo get_network_option( 1, "mykey" );'`
+    Then STDOUT should be:
+      """
+      456
+      """
--- a/features/network-meta.feature
+++ b/features/network-meta.feature
@@ -51,7 +51,7 @@ Feature: Manage network-wide custom fields.
 
     When I run `wp eval 'update_network_option( 1, "objkey", "123" );'`
 
-    When I run `wp network meta get 1 objkey`
+    And I run `wp network meta get 1 objkey`
     Then STDOUT should be:
       """
       123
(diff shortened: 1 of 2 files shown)


spacedmonkey:
Pinging maintainers to get eyes on this @swissspidy @danielbachhuber @schlessera

swissspidy:
Note that we get pinged automatically with the review request :)

swissspidy:
Two thoughts:

1. #504 mentions steps to reproduce the issue — this could be added as a Behat test.
2. Looking at https://core.trac.wordpress.org/ticket/61467, it sounds like this could just as well be fixed only in core — so why this change here? Could we just wait for the core fix?

jonnynews:
> Note that we get pinged automatically with the review request :)

No one was assigned for a code reviewers. Maybe this repo needs a codeowner file 🤔 

> https://github.com/wp-cli/entity-command/issues/504 mentions steps to reproduce the issue — this could be added as a Behat test.

Before spending any time on getting tests to work, I wanted to make sure everyone was happy with the approach. 

> Looking at https://core.trac.wordpress.org/ticket/61467, it sounds like this could just as well be fixed only in core — so why this change here? Could we just wait for the core fix?

The core patch is not confirmed yet and not work. Also that will only fix the issue with 6.8+. WP CLI support down to 3.7. Fixing this here is to ensure BC.

swissspidy:
> No one was assigned for a code reviewers. 

Screenshot 2025-01-22 at 14 51 12
Screenshot 2025-01-22 at 14 51 05

> Before spending any time on getting tests to work, I wanted to make sure everyone was happy with the approach.

It looks reasonable to me in theory, yes.

jonnynews:
> > No one was assigned for a code reviewers.

This is what I see

Screenshot 2025-01-22 at 09 06 11

schlessera:
Yes, approach looks good, @jonnynews. If we add the tests from #504 , this will be good to go.

swissspidy:
Turns out, there is actually 1 failing test in extension-command that will be fixed by this change.

https://github.com/wp-cli/extension-command/blob/252674e3fb310e1a6cbb946cfb15d70c6cdabc63/features/theme.feature#L319-L394

```
001 Scenario: Enabling and disabling a theme # features/theme.feature:320
      Then STDOUT should contain:            # features/theme.feature:377
        $ wp network-meta get 1 allowedthemes
        array (
          'twentytwentyfive' => true,
        )
        
        cwd: /tmp/wp-cli-test-run--69c497154be549.74820243/
        run time: 0.33636903762817
        exit status: 0
        
        Output does not contain expected string:
        'moina-blog' => true
        
        --- Expected
        +++ Actual
        @@ @@
        -'moina-blog' => true
        +array (
        +  'twentytwentyfive' => true,
        +)
         (Exception)
```

As you can see, it uses `wp network-meta`.

The test passes when using this PR.

Though it also passes when replacing `wp network-meta` with `wp site option`

swissspidy (approved):

features/network-meta.feature
```diff
@@ -16,3 +16,69 @@ Feature: Manage network-wide custom fields.
...
+      456
+      """
+
+  @require-object-cache
+  Scenario: Object cache correctly handles network meta updates
+    Given a WP multisite install
+
+    When I run `wp eval 'update_network_option( 1, "objkey", "123" );'`
+
+    When I run `wp network meta get 1 objkey`
```
swissspidy:
```suggestion
    And I run `wp network meta get 1 objkey`
```

features/network-meta.feature
```diff
@@ -16,3 +16,69 @@ Feature: Manage network-wide custom fields.
...
+    Then STDOUT should be:
+      """
+      123
+      """
+
+    When I run `wp eval 'update_network_option( 1, "objkey", "456" );'`
+
+    When I run `wp network meta get 1 objkey`
```
swissspidy:
```suggestion
    And I run `wp network meta get 1 objkey`
```

Issue #519, opened 2025-01-30
support `WP_Query`'s `date_query` values from `wp post list` command-line


drzraf:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

`wp post list` support arbitrary `--<field>` being passed to `WP_Query` but [date parameters](https://developer.wordpress.org/reference/classes/wp_query/#date-parameters) which depends on a nested structure are not handled. Doing so would allow handy filtering logic based on post date attributes. 

**Describe the solution you'd like**

`--date-<field>` could be used as a prefix opening the way to things like `--date-before=2024-01-01` or `--date-year=2025`
(The same applied to `meta_query` would also be great as it would resolves the long-standing problem of retrieving posts based on meta-condition straight from `wp post list` without any piping/subcommand)

swissspidy:
Thanks for the suggestion!

Perhaps we could support JSON input for these cases.

swissspidy:
Previously:

* https://github.com/wp-cli/wp-cli/issues/3039
* https://github.com/wp-cli/wp-cli/issues/4616

`\WP_CLI\Utils\parse_shell_arrays` is the easy solution here.

drzraf:
Could you please add `tax_query` to this `$array_arguments` variable. I tested, it works well!

swissspidy:
Are you perhaps up for submitting a PR? :-)

drzraf:
Yes... but... I fear... you may legitimately ask for behat tests and their
> When I run `wp term create post_tag foo`
> and when I run `wp post create --tax_input=stuff`
> and    🙄⏳🫣⏳🤯

swissspidy:
See, it basically writes itself :-)

But here you go: #522

PR #522, opened 2025-02-21
`wp post list`: Add JSON input support for `tax_query` and `meta_query`

swissspidy:
Follow-up to #520, see #519

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -447,6 +447,12 @@ Feature: Manage WordPress posts
       | {POST_ID} | key2     | value2b    |
       | {POST_ID} | key3     | value3     |
 
+    When I run `wp post list --field=post_title --meta_query='[{"key":"key2","value":"value2b"}]'`
+    Then STDOUT should contain:
+      """
+      Test Post
+      """
+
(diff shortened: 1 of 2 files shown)


renatonascalves (approved):
🌵

features/post.feature
```diff
@@ -365,18 +365,38 @@ Feature: Manage WordPress posts
...
       new post
       """
 
+  Scenario: List posts with tax query
+    When I run `wp term create category "First Category" --porcelain`
+    And save STDOUT as {TERM_ID}
+    And I run `wp term create category "Second Category" --porcelain`
+    And save STDOUT as {SECOND_TERM_ID}
+
+    When I run `wp post create --post_title='post-1' --post_category="First Category"`
```
mrsdizzie, line 387: Was this meant to be ```{TERM_ID}``` from above? (similar for next line)? I think its the same either way since it takes an ID or name, but other test uses ```{TERM_ID}``` and I don't see it used here otherwise.

swissspidy: Removed now

mrsdizzie (approved):

PR #523, opened 2025-02-26
Add a flag for getting post meta as a single value or not

### LINKED ISSUE #325: Support non unique meta_keys when getting values
## Feature Request

Post meta can either be unique or non single. This means that the same meta key can be used multiple times in post meta. When using `get_post_meta()` you can pass if the meta key is single or not.

Unfortunately `wp post meta 1 meta_key` does not support this. At the moment only the first value is returned. 

This is caused by the fixed defined parameter in the code: https://github.com/wp-cli/entity-command/blob/6e0e77a297eefa3329b94bec16c15cf7528d343f/src/WP_CLI/CommandWithMeta.php#L161

It would be nice if the third parameter could be either set using the arguments or using a flag. Either
`wp post meta 1 meta_key false` or `wp post meta 1 meta_key --is-single=false`.

BhargavBhandari90:
Fixes #325 

- Currently marked as WIP.
- I haven't wrote behat tests for this yet.
- Let's finalize the changes first and then I'll write the tests.

**What I did**
- Added a flag `--is-single`
- Default value will be 1 ( true ) which was already there.
- And we can now pass 0 as false for getting multiple values

**Command Examples**

- `wp post meta get 1 bwp_key --is-single=0`
- `wp post meta get 1 bwp_key --is-single=1`

### DIFF
--- a/features/post-meta.feature
+++ b/features/post-meta.feature
@@ -219,6 +219,39 @@ Feature: Manage post custom fields
       My\New\Meta
       """
 
+Scenario: List post meta with or without single flag
+    Given a WP install
+
+    When I run `wp post meta add 1 apple banana`
+    And I run `wp post meta add 1 apple banana`
+    Then STDOUT should not be empty
+
+    When I run `wp post meta get 1 apple`
+    Then STDOUT should be:
+      """
+      banana
+      """
+
+    When I run `wp post meta get 1 apple --single`
+    Then STDOUT should be:
+      """
+      banana
+      """
+
+    When I run `wp post meta get 1 apple --no-single`
+    Then STDOUT should be:
+      """
+      array (
+        0 => 'banana',
+        1 => 'banana',
+      )
+      """
+    When I run `wp post meta get 1 apple --no-single --format=json`
+    Then STDOUT should be:
+      """
+      ["banana","banana"]
+      """
+
   @pluck
   Scenario: Nested values can be retrieved.
     Given a WP install
--- a/features/post-meta.feature
+++ b/features/post-meta.feature
@@ -223,7 +223,7 @@ Scenario: List post meta with or without single flag
     Given a WP install
 
     When I run `wp post meta add 1 apple banana`
-    And I run the previous command again
+    And I run `wp post meta add 1 apple mango`
     Then STDOUT should not be empty
 
     When I run `wp post meta get 1 apple`
@@ -243,13 +243,13 @@ Scenario: List post meta with or without single flag
       """
       array (
         0 => 'banana',
-        1 => 'banana',
+        1 => 'mango',
       )
       """
     When I run `wp post meta get 1 apple --no-single --format=json`
     Then STDOUT should be:
       """
-      ["banana","banana"]
+      ["banana","mango"]
       """
 
   @pluck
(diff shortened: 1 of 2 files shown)


src/WP_CLI/CommandWithMeta.php
```diff
@@ -142,6 +142,15 @@ function ( $a, $b ) use ( $orderby, $order ) {
...
 	 * : The name of the meta field to get.
 	 *
+	 * [--is-single=<is-single>]
+	 * : Whether to return a single value. This parameter has no effect if $key is not specified.
+	 * ---
+	 * default: 1
+	 * options:
+	 *   - 1
+	 *   - 0
+	 * ---
```
swissspidy, line 152: We don't typically use such a format in WP-CLI for boolean flags. Instead we use just `--is-single` for "true", and don't pass a flag if you want "false" (or `--no-is-single`, which is automatically supported)

So this would be simply:

```
* [--is-single]
* : Whether to return a single value. This parameter has no effect if $key is not specified.
```

And then you can use `$is_single = Utils\get_flag_value( $assoc_args, 'is-single', true );`

Also, I'd probably rename to just `--single`

BhargavBhandari90: Got it. Making changes. Thank you for your feedback.

BhargavBhandari90:
@swissspidy After your feedback, now command will be 

- wp post meta get 1 bwp_key --single // For true
- wp post meta get 1 bwp_key --no-single // For false

And without the `single` flag it will work the way it was working.

swissspidy:
Sounds good! Next steps: tests.

BhargavBhandari90:
@swissspidy  Yep. On it

features/post-meta.feature
```diff
@@ -219,6 +219,39 @@ Feature: Manage post custom fields
       My\New\Meta
       """
 
+Scenario: List post meta with or without single flag
+    Given a WP install
+
+    When I run `wp post meta add 1 apple banana`
+    And I run `wp post meta add 1 apple banana`
```
ernilambar, line 226: This `And` line seems redundant.

BhargavBhandari90: Should we use `And I run the previous command again`?

src/WP_CLI/CommandWithMeta.php
```diff
@@ -142,6 +142,9 @@ function ( $a, $b ) use ( $orderby, $order ) {
 	 * <key>
 	 * : The name of the meta field to get.
 	 *
+	 * [--single]
+	 * : Whether to return a single value. This parameter has no effect if $key is not specified.
```
ernilambar, line 146: <key> is required here in this command. I think command wont even run without that argument. Not sure second sentence is needed here. What do you think?

BhargavBhandari90: I've updated the documentation (Removed 2nd sentence).

I tried the command without `--single` and it is working fin without it. I thin `--` makes it optional argument. 
image

features/post-meta.feature
swissspidy: For testing purposes here I would add multiple different meta values, not just banana

swissspidy (approved):

PR #527, opened 2025-03-18
Ensure wp user update Returns Non-Zero Exit Code for Invalid Users

karthick-murugan:
Fixes https://github.com/wp-cli/wp-cli/issues/5960

This PR fixes the issue where `wp user update` incorrectly exits with **0** when the provided user ID is invalid. Now, when an invalid user ID, email, or login is supplied, the command will return an exit code of **1** instead of **0**, ensuring proper error handling.

**Testing Instructions**

**Try updating a non-existent user:**
`wp user update 9999 --user_pass=securepassword`

**Expected output:**
`Warning: Invalid user ID, email or login: '9999'`

**Expected exit code: 1**
`echo $?`
_Should return 1_

**Try updating a existent user:**
`wp user update 1 --user_pass=securepassword`

**Expected output:**
`Success: Updated user 1.`

**Expected exit code: 0**
`echo $?`
_Should return 0_

**Screenshot**

user_update_command

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -147,6 +147,13 @@ Feature: Manage WordPress users
       3
       """
 
+    When I try `wp user update 9999 --user_pass=securepassword`
+    Then the return code should be 1
+    And STDERR should contain:
+      """
+      Error: No valid users found.
+      """
+
   Scenario: Delete user with invalid reassign
     Given a WP install
     And a session_no file:
@@ -728,12 +735,3 @@ Feature: Manage WordPress users
       """
       newtestuser
       """
-  Scenario: Updating an invalid user should return an error
-    Given a WP install
-    When I try `wp user update 9999 --user_pass=securepassword`
-    Then the return code should be 1
-    And STDERR should contain:
-      """
-      Error: No valid users found.
-      """
-
--- a/features/user.feature
+++ b/features/user.feature
@@ -735,3 +735,4 @@ Feature: Manage WordPress users
       """
       newtestuser
       """
+
--- a/features/user.feature
+++ b/features/user.feature
@@ -728,4 +728,12 @@ Feature: Manage WordPress users
       """
       newtestuser
       """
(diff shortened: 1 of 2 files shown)


src/User_Command.php
```diff
@@ -545,6 +545,10 @@ public function update( $args, $assoc_args ) {
 			$user_ids[] = $user->ID;
 		}
 
+		if ( empty( $user_ids ) ) {
+			exit( 1 );
```
mrsdizzie, line 549:
```suggestion
			WP_CLI::error( 'No valid users found.' )
```
If we are exiting unsuccessfully I think we should give a reason and let the user know it is intentional. Otherwise it seems like we either die on warnings, or that something has gone wrong since there is no other output. We can't error on the invalid user itself because there might be more than one, but we can here.

mrsdizzie (changes requested):
Also we should add a test for this : )

src/User_Command.php
karthick-murugan: @mrsdizzie - Thank you reviewing this PR. The requested changes has been completed. Please have a look.

features/user.feature
```diff
@@ -728,4 +728,12 @@ Feature: Manage WordPress users
       """
       newtestuser
       """
+  Scenario: Updating an invalid user should return an error
+    Given a WP install
+    When I try `wp user update 9999 --user_pass=securepassword`
```
mrsdizzie, line 733: Every time we use `Given a WP install` that is a new copy of WordPress the tests need to set up which takes more time etc... This test can go inside the existing `Scenario: User CRUD operations` since it fits there and doesn't require a unique state.

mrsdizzie (changes requested):

features/user.feature
karthick-murugan: @mrsdizzie - Updated the changes

mrsdizzie (approved):
Looks good, thanks!

PR #528, opened 2025-03-20
Validate Parameter Names in wp user update Command

### LINKED ISSUE #5286: wp user update silently ignores typo'ed --user_pass flag
With WP-CLI 2.3.0, I accidentally typo'ed the --user_pass param to "wp user update" as --user-pass (i.e. a minus instead of an underscore). Here's what was output:

wp user update myuser --user-pass=mypass
Success: Updated user 1.

However, it didn't set "mypass" as the user password (it left it as the old one) and didn't report that --user-pass isn't a valid flag. When I used the correct --user_pass flag, I got the same success message but it did change the password to mypass. It needs to give an error message about --user-pass being a bad flag instead of a success message. I didn't check if this applies to similar typos with other user update flags (many of them have underscores).

karthick-murugan:
Fixes https://github.com/wp-cli/wp-cli/issues/5286

Summary

This PR addresses issue [#5286](https://github.com/wp-cli/wp-cli/issues/5286), where `wp user update `silently ignored incorrect parameter names (e.g., `--user-pass` instead of `--user_pass`) without providing an error message.

With this fix, WP-CLI will now validate parameter names and issue a warning when an unknown parameter is used.
Changes Introduced

 Added validation to check whether the parameters passed to `wp user update` match the expected field names from `wp_update_user().` If an unrecognized parameter is provided, WP-CLI will display a warning and prevent the update from proceeding.

**Behavior Before Fix**

```
$ wp user update 6 --user-pass=mypass  
Success: Updated user 6.  
```

(Note: The password was not updated, and no warning was shown.)

**Behavior After Fix**

```
$ wp user update 6 --user-pass=mypass  
Warning: Unknown --user-pass parameter.  
Error: Need some fields to update.  
```

(This prevents silent failures and informs the user of incorrect parameter usage.)

**How to Test**

1. Setup a WordPress environment with WP-CLI installed.
2. Create a test user (if not already available):

`wp user create testuser test@example.com --role=subscriber --user_pass=oldpass`

3. Run a valid update command:

`wp user update testuser --user_pass=newpass  `

✅ Expected: User password should update successfully.

4. Run an update command with a typo:

`wp user update testuser --user-pass=newpass  `

**❌ Expected:**

```
Warning: Unknown --user-pass parameter.  
Error: Need some fields to update.  
```

(This ensures that incorrect parameters are flagged.)

5. Test with another incorrect parameter:

`wp user update testuser --wrong_flag=value  `

❌ Expected: Similar warning about an unknown parameter.

6. Verify that no changes were made when incorrect parameters were used.

**Screenshots**

user_update_params

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -376,22 +376,10 @@ Feature: Manage WordPress users
       administrator
       """
 
-    And a user exists with login 9999
-
-    When I run `wp user update 9999 --user-pass=admin`
-    Then STDERR should contain:
-      """
-      Error: Parameter errors:
-      unknown --user-pass parameter
-      Did you mean '--user_pass'?
-      """
-
-    And a user exists with login 9999
-    
-    When I run `wp user update 9999 --user_pass=admin`
+    When I run `wp user update 1 --user_pass=new_password`
     Then STDOUT should contain:
       """
-      Success: Updated user 9999.
+      Success: Updated user 1.
       """
 
   Scenario: Managing user capabilities
--- a/features/user.feature
+++ b/features/user.feature
@@ -376,6 +376,24 @@ Feature: Manage WordPress users
       administrator
       """
 
+    And a user exists with login 9999
+
+    When I run `wp user update 9999 --user-pass=admin`
+    Then STDERR should contain:
+      """
+      Error: Parameter errors:
+      unknown --user-pass parameter
+      Did you mean '--user_pass'?
+      """
+
+    And a user exists with login 9999
+    
+    When I run `wp user update 9999 --user_pass=admin`
+    Then STDOUT should contain:
+      """
+      Success: Updated user 9999.
+      """
+
   Scenario: Managing user capabilities
     Given a WP install
 
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -536,6 +536,49 @@ class User_Command extends CommandWithDBObject {
 			unset( $assoc_args['user_login'] );
 		}
 
+		// Define valid fields for wp_update_user().
+		$valid_fields = array(
+			'user_pass',
+			'user_nicename',
+			'user_url',
+			'user_email',
+			'display_name',
+			'nickname',
+			'first_name',
+			'last_name',
+			'description',
+			'rich_editing',
+			'user_registered',
+			'role',
+			'syntax_highlighting',
+			'comment_shortcuts',
+			'admin_color',
+			'use_ssl',
+			'show_admin_bar_front',
+			'locale',
+		);
+
+		// Check for invalid field names.
+		$invalid_fields = array();
+		foreach ( array_keys( $assoc_args ) as $field ) {
+			// Skip special flags like 'skip-email'.
+			if ( 'skip-email' === $field ) {
+				continue;
+			}
+
+			if ( ! in_array( $field, $valid_fields, true ) ) {
+				$invalid_fields[] = $field;
+			}
+		}
+
+		// Warn about invalid fields.
+		if ( ! empty( $invalid_fields ) ) {
+			foreach ( $invalid_fields as $field ) {
+				WP_CLI::warning( "Unknown --{$field} parameter." );
+				unset( $assoc_args[ $field ] );
+			}
+		}
+
 		if ( isset( $assoc_args['role'] ) ) {
 			self::validate_role( $assoc_args['role'], true );
 		}
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -518,9 +518,6 @@ class User_Command extends CommandWithDBObject {
 	 * [--role=<role>]
 	 * : A string used to set the user's role.
 	 *
-	 * --<field>=<value>
-	 * : One or more fields to update. For accepted fields, see wp_update_user().
-	 *
 	 * [--skip-email]
 	 * : Don't send an email notification to the user.
 	 *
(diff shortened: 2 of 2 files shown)


mrsdizzie:
Thank you for the PR! This needs a test to verify any new behavior.

That being said, I'm not totally sure if this is the right approach. If the goal is to accept any field and pass them to `wp_update_user`, It should be the job of that function to return this error. Unfortunately, after a bit of quick testing it looks like `wp_update_user` doesn't return an error when you pass it a random value, so I guess relying on that isn't possible : (

If that isn't possible, it seems better to just keep all of the valid options in the DocBlock, not accept arbitrary `--field=<value>` and use the existing wp cli parameter validation which already has code to handle wrong arguments like this. This happens automatically if you remove this code from the doc bloc:
```php
/**
* --<field>=<value>
* : One or more fields to update. For accepted fields, see wp_update_user().
*
**/
```
And even better, it gets built in support for similar commands. So if we removed that, you would then see this:
```sh
$ wp user update 1 --user-pass=test --user-email=test@example.com
Error: Parameter errors:
 unknown --user-pass parameter
Did you mean '--user_pass'?
 unknown --user-email parameter
Did you mean '--user_email'?
```

Then all of the fields would be forced to have to have documentation as well, which seems like a benefit. It doesn't feel like it makes sense to keep some of the accepted fields in the DocBlock, then say we accept other arbitrary fields (not documented), but then not accept them based on a second hardcoded list. I think it would be better to just keep them all in the DocBlock and get the advantages above.

Maybe other maintainers have thoughts.

swissspidy:
FWIW, this was the original suggestion on the mentioned issue, which wasn‘t addressed here:

> Yes, I agree, WP-CLI should verify the flag against a whitelist. This would need to be kept in sync with the WordPress Core version, though, and we need to deal with the changes between Core versions as well.
> 
> I'm thinking about an external list of entities and their properties, with starting and/or ending version numbers for each. Having that outside of the individual commands would make it easier to maintain it and check it against new releases.

Some properties were only introduced in a specific WordPress version, or might have been removed again (though that seems unlikely to me, because of back compat), so I get where this is coming from. We should at least discuss the fessibility of this more. NB this goes not only for users but also other entities

I do like the suggestion with the docblock though, because it makes for a much nicer user experience. But we probably can‘t combine that with such a dynamic allowlist…

That said, maybe plugins can add support for more properties too, at which point the allowlist would be too limiting.

karthick-murugan:
Thanks for the feedback, @mrsdizzie @swissspidy !

I've updated the DocBlock to list all accepted fields and removed --<field>=<value> validation, letting WP-CLI handle errors automatically. 

Regarding the dynamic allowlist, I see the value in keeping it in sync with WordPress Core, but it could add maintenance overhead and limit plugin flexibility. We should discuss the feasibility of this further—not just for users, but for other entities as well.

Looking forward to more thoughts! 🚀

mrsdizzie:
@swissspidy I agree I'm not 100% convinced it feels right to filter properties here, though it seems the trade off is a worse user experience for wp cli users. I do think ideally this would be the responsibility of the `wp user update` function or something in core. I am also not sure if plugins can/do change this behavior, but if so it would be another reason not to do this here.

The suggested path in the issue feels complicated to me given the seemingly [small amount of changes](https://developer.wordpress.org/reference/functions/wp_insert_user/#changelog) upstream to what they accept. And it wouldn't really handle the case of plugins being able to alter what is accepted (as you say, no list can).

I think a single line in a docblock that says `*WordPress 4.7 and later feels` 'good enough' vs an entire abstraction system.

swissspidy:
> I am also not sure if plugins can/do change this behavior, but if so it would be another reason not to do this here.

I thought maybe via the `wp_pre_insert_user_data` filter, but couldn't find any plugins doing this.

So yeah, perhaps having a thorough docblock is good enough.

We already list some fields in the docblock, but not all of them. So expanding that list would be helpful either way.

An alternative/additional solution could be to provide suggestions for typos like in the reported `--user-pass` vs `--user_pass` use case. Why doesn't this result in an "Did you mean --user_pass" message?

mrsdizzie:
> An alternative/additional solution could be to provide suggestions for typos like in the reported `--user-pass` vs `--user_pass` use case. Why doesn't this result in an "Did you mean --user_pass" message?

This happens automatically via WP CLI when validating what the user passed against what the command supports (via the DockBlock) but that check never happens when a command has
```
--<field>=<value>
```
Because the command accepts *anything* as valid, so it doesn't run any validation code. Removing that provides the validation / suggestion at the expense of not allowing arbitrary fields anymore -- which is probably what is best in this specific example (and to make sure the command knows and documents all of the accepted values).

swissspidy:
> Because the command accepts anything as valid, so it doesn't run any validation code. 

Ah, I see, that's it. I wonder if we could still do validation but then only emit a warning instead of an error. But then one probably would need a way to ignore the warning...

So yeah, I suppose starting with docblock improvements is the best next step.

Here are all occurrences of `--<field>=<value>` across all repos btw: https://github.com/search?q=org%3Awp-cli+%22--%3Cfield%3E%3D%3Cvalue%3E%22+path%3Aphp&type=code

and in entity-command alone: https://github.com/search?q=repo%3Awp-cli%2Fentity-command+%22--%3Cfield%3E%3D%3Cvalue%3E%22+language%3APHP&type=code&l=PHP

Lot of room for improvement there IMO.

mrsdizzie:
I think the problem with a warning is that you still have the same initial confusion that `wp_update_user` will return success. So you'd end up with:

```
wp user update 1 --user-pass=test
Warning: unknown --user-pass parameter
Did you mean '--user_pass'?
Success: Updated user 1.
```

It isn't really clear what is happening there -- is the warning wrong? Did it update something? You also don't end up with an error code and wouldn't see the warning if using --quiet (I think?). I think an error would be better for the user.

swissspidy:
Valid point 👍

karthick-murugan:
@mrsdizzie - When there is invalid field mentioned, there is only Error message and not a success message. Please have a look at this screenshot.

image

swissspidy:
Yeah I know, but my idea was a _warning_, in which case you would see both. But let's discard the warning idea :)

I think I'd be fine with more thorough documentation in favor of the dynamic catch-all `--<field>=<value>` bit. But we would need to do this for all commands, not just users.

karthick-murugan:
@swissspidy To address our main issue and improve parameter validation across WP-CLI commands, these are the changes proposed. Can I proceed with these?

Remove --<field>=<value> syntax from the DocBlock of wp user update **( which is already done )** and replace it with listed parameters (e.g., --user_pass, --user_email, etc.). 

Extend this pattern to other commands (e.g., post update, term update) in follow-up PRs, replacing dynamic field syntax with documented parameters.

PR #529, opened 2025-04-02
Fix Gherkin lint issues

ernilambar:


ernilambar:
Issue of Gherkin lint should be fixed after https://github.com/wp-cli/.github/pull/128 is merged. Other tests are green.

PR #530, opened 2025-04-05
Improve wp remove-cap command

### LINKED ISSUE #98: Passing a role name to `wp user remove-cap` should probably return an error
At the user level, WordPress internally treats roles and capabilities the same. They're all stored in the `wp_capabilities` user meta field. This is unfortunate.

This means that `wp user remove-cap <user> <role>` works, which can have unexpected consequences if, for example, a script blindly loops over capabilities in order to remove them from a user and one of the capabilities is a role name.

I think if an attempt is made to remove a capability from a user that matches a role name, it should return an error along with recommending that `wp user remove-role` is used instead.

BhargavBhandari90:
Fixes #98 

---

Let's start with this:

---

- Added Role `ExMember` and capability `exmember`
- Now when we try to run comman: `wp user remove-cap 1 exmember`
- It will have error ( see the screenshot below ): There is a role similar to 'exmember' capability. Use `wp user remove-role` instead.
image

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -392,7 +392,7 @@ Feature: Manage WordPress users
       contributor
       """
 
-    When I run `wp user remove-cap {USER_ID}` contributor
+    When I run `wp user remove-cap {USER_ID} contributor`
     Then the return code should be 1
     And STDERR should be:
       """
--- a/features/user.feature
+++ b/features/user.feature
@@ -379,26 +379,6 @@ Feature: Manage WordPress users
       administrator
       """
 
-  Scenario: Show error when trying to remove capability same as role
-    Given a WP install
-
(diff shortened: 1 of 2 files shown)


swissspidy:
Thanks for starting this!

The example you mentioned would make for a good Behat test :-) Would you be up for adding one?

BhargavBhandari90:
Sure @swissspidy I'll add the Behat test. Thank you.

BhargavBhandari90:
@swissspidy added behat test. Let me know if this works.

swissspidy:
The original suggestion in https://github.com/wp-cli/entity-command/issues/98#issuecomment-339877391 was to add a `--force` flag to ignore the error and still remove the cap, maintaining existing behavior

schlessera (approved):

PR #532, opened 2025-04-11
`wp post create`: Add JSON input support for `tax_input`

caribe:
Fixes this 10 year old issue https://github.com/wp-cli/wp-cli/issues/1323 (closed, but not fixed).

`wp post create` parameter `--tax_input` is supposed to be an array, but no JSON parsing is done and therefore is unusable.

### DIFF
--- a/src/Post_Command.php
+++ b/src/Post_Command.php
@@ -186,7 +186,7 @@ class Post_Command extends CommandWithDBObject {
 			WP_CLI::warning( "The 'meta_input' field was only introduced in WordPress 4.4 so will have no effect." );
 		}
 
-		$array_arguments = [ 'meta_input' ];
+		$array_arguments = [ 'meta_input', 'tax_input' ];
 		$assoc_args      = Utils\parse_shell_arrays( $assoc_args, $array_arguments );
 
 		if ( isset( $assoc_args['from-post'] ) ) {


swissspidy:
Thanks for opening this PR.

Would you also be up for adding some tests to cover this change? See #133 for inspiration.

swissspidy:
I just added a test now, but it fails because `tax_input` requires the current user (which there technically isn't one with WP-CLI) to have the necessary capabilities. So we need to figure that one out first...

Related: https://github.com/orgs/wp-cli/discussions/6052

features/post.feature
```diff
@@ -395,6 +395,25 @@ Feature: Manage WordPress posts
...
       """
 
+  Scenario: Creating/updating posts with taxonomies
+    When I run `wp term create category "First Category" --porcelain`
+    And save STDOUT as {CAT_1}
+    And I run `wp term create category "Second Category" --porcelain`
+    And save STDOUT as {CAT_2}
+    And I run `wp term create post_tag "Term One" --porcelain`
+    And I run `wp term create post_tag "Term Two" --porcelain`
+    And I run `wp post create --post_title='Test Post' --post_content='Test post content' --tax_input='{"category":[{CAT_1},{CAT_2}],"post_tag":["term-one", "term-two"]}' --porcelain`
```
seanlanglands, line 432: @swissspidy, in this scenario, would passing `--user=admin` in this step be acceptable in your opinion?

```suggestion
    And I run `wp post create --post_title='Test Post' --post_content='Test post content' --tax_input='{"category":[{CAT_1},{CAT_2}],"post_tag":["term-one", "term-two"]}' --user=admin --porcelain`
```

This user is the default administrator when core is installed during testing:
https://github.com/wp-cli/wp-cli-tests/blob/5b42c271c96efd1072cecaa8b2923db52318b056/src/Context/FeatureContext.php#L1392

swissspidy: That would fix the test but not the user experience in general.

We would need to document that `tax_input` only works in combination with `--user`, which is odd.

Another option could be to temporarily filter `user_has_cap` for this particular code path so that it would always work. But not sure if there's precedence in WP-CLI.

Perhaps @schlessera has a good idea.

schlessera: You can find some context in the discussion here: https://github.com/wp-cli/entity-command/pull/207

This is a bug in the capabilities configuration of WordPress Core, and it basically means that `tax_input` (which is supposed to be generic) will ONLY work for custom taxonomies, but not for the built-in ones. So if you switch from `category` to a custom taxonomy, it should work without administrator access.

schlessera: Initially, I had thought about adding a custom work-around in WP-CLI, but also wanted to have this flagged as a bug in WP Core. Not sure what the latest status is for Core.
But I'm happy to consider a workaround. At the time, I thought that `tax_input` should defer to `post_tag` or `post_catgory` for those, but I can't remember how involved that would have been.

swissspidy: Added a workaround now & some documentation

PR #536, opened 2025-05-18
PHPStan level 9

swissspidy:
Blocked by:

* https://github.com/wp-cli/wp-cli/pull/6096
* https://github.com/wp-cli/wp-cli-tests/pull/262

### DIFF
--- a/features/user.feature
+++ b/features/user.feature
@@ -190,7 +190,7 @@ Feature: Manage WordPress users
       --reassign parameter is invalid. All associated posts will be deleted. Proceed? [y/n]
       """
 
-    When I try `wp user delete bobjones < session_yes`
+    When I run `wp user delete bobjones < session_yes`
     And I run `wp post list --format=count`
     Then STDOUT should be:
       """
--- a/features/user.feature
+++ b/features/user.feature
@@ -190,7 +190,7 @@ Feature: Manage WordPress users
       --reassign parameter is invalid. All associated posts will be deleted. Proceed? [y/n]
       """
 
-    When I run `wp user delete bobjones < session_yes`
+    When I try `wp user delete bobjones < session_yes`
     And I run `wp post list --format=count`
     Then STDOUT should be:
       """
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-  level: 6
+  level: 9
   paths:
     - src
     - entity-command.php
@@ -9,10 +9,6 @@ parameters:
   scanFiles:
     - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
   treatPhpDocTypesAsCertain: false
-  dynamicConstantNames:
-  	- WP_DEBUG
-  	- WP_DEBUG_LOG
-  	- WP_DEBUG_DISPLAY
   ignoreErrors:
     - identifier: missingType.iterableValue
     - identifier: missingType.property
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -5,7 +5,6 @@ parameters:
     - entity-command.php
   scanDirectories:
     - vendor/wp-cli/wp-cli/php
-    - vendor/wp-cli/wp-cli-tests
   scanFiles:
     - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
   treatPhpDocTypesAsCertain: false
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,0 +1,20 @@
+parameters:
+  level: 6
+  paths:
+    - src
+    - entity-command.php
+  scanDirectories:
+    - vendor/wp-cli/wp-cli/php
+    - vendor/wp-cli/wp-cli-tests
+  scanFiles:
+    - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
+  treatPhpDocTypesAsCertain: false
+  dynamicConstantNames:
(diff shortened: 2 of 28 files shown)


src/WP_CLI/CommandWithDBObject.php
```diff
@@ -184,17 +184,4 @@ protected function get_formatter( &$assoc_args ) {
...
-
-	/**
-	 * Given a callback, display the URL for one or more objects.
-	 *
-	 * @param array $args One or more object references.
-	 * @param string $callback Function to get URL for the object.
-	 */
-	protected function _url( $args, $callback ) {
-		foreach ( $args as $obj_id ) {
-			$object = $this->fetcher->get_check( $obj_id );
```
swissspidy, line 196: The `fetcher` property didn't even exist on this abstract class, and in the extending classes it was `private`. So this code never worked. Plus it was never used, so it's safe to remove.

src/Post_Command.php
```diff
@@ -182,10 +182,6 @@ public function create( $args, $assoc_args ) {
 			$assoc_args['post_category'] = $this->get_category_ids( $assoc_args['post_category'] );
 		}
 
-		if ( isset( $assoc_args['meta_input'] ) && Utils\wp_version_compare( '4.4', '<' ) ) {
```
swissspidy, line 185: Figured with our move to WP 4.9+ we can remove this compat code.

src/Option_Command.php
```diff
@@ -321,8 +322,8 @@ public function list_( $args, $assoc_args ) {
 				function ( $a, $b ) use ( $orderby, $order ) {
 					// Sort array.
 					return 'asc' === $order
-						? $a->$orderby > $b->$orderby
-						: $a->$orderby < $b->$orderby;
+						? $a->$orderby <=> $b->$orderby
+						: $b->$orderby <=> $a->$orderby;
```
swissspidy, line 326: I hope this is not the wrong way around 🤔

phpstan.neon.dist
```diff
@@ -0,0 +1,16 @@
+parameters:
+  level: 9
+  paths:
+    - src
+    - entity-command.php
+  scanDirectories:
+    - vendor/wp-cli/wp-cli/php
+    - vendor/wp-cli/wp-cli-tests
```
swissspidy:
```suggestion
```

features/user.feature
```diff
@@ -190,7 +190,7 @@ Feature: Manage WordPress users
       --reassign parameter is invalid. All associated posts will be deleted. Proceed? [y/n]
       """
 
-    When I run `wp user delete bobjones < session_yes`
+    When I try `wp user delete bobjones < session_yes`
```
schlessera: Why was this change needed?

src/Menu_Item_Command.php
```diff
@@ -92,7 +92,7 @@ class Menu_Item_Command extends WP_CLI_Command {
 	public function list_( $args, $assoc_args ) {
 
 		$items = wp_get_nav_menu_items( $args[0] );
-		if ( false === $items || is_wp_error( $items ) ) {
+		if ( false === $items ) {
```
schlessera, line 95: Is it impossible here to end up with a `WP_Error`? Relying on the docblocks is not enough, and I'm seeing that multiple filters are involved... 🤔

src/Menu_Item_Command.php
```diff
@@ -408,10 +410,10 @@ public function delete( $args, $assoc_args ) {
 	private function add_or_update_item( $method, $type, $args, $assoc_args ) {
 
 		$menu            = $args[0];
-		$menu_item_db_id = Utils\get_flag_value( $args, 1, 0 );
+		$menu_item_db_id = $args[1] ?? 0;
 
 		$menu = wp_get_nav_menu_object( $menu );
-		if ( ! $menu || is_wp_error( $menu ) ) {
+		if ( false === $menu ) {
```
schlessera, line 416: Same here re. `WP_Error`. I'm assuming there was a reason why that particular check was in there?

src/Menu_Location_Command.php
```diff
@@ -147,18 +149,20 @@ public function assign( $args, $assoc_args ) {
...
 	 * @subcommand remove
+	 *
+	 * @param array{string, string} $args
 	 */
 	public function remove( $args, $assoc_args ) {
 
 		list( $menu, $location ) = $args;
 
 		$menu = wp_get_nav_menu_object( $menu );
-		if ( ! $menu || is_wp_error( $menu ) ) {
```
schlessera, line 156: Same here re. `WP_Error`.

src/Option_Command.php
schlessera: Looks correct to me.

schlessera (dismissed):

features/user.feature
swissspidy: Whoops, probably simply by accident

src/Menu_Item_Command.php
swissspidy: There is no code path in there that returns WP_Error, I already checked.

Filters can always be abused, but we can‘t check for every possible return value.

git blame was unfortunately not helpful here (mangled history)

src/Menu_Item_Command.php
swissspidy: I don't think so.

It was originally introduced in 246a8bc13c55f26c81f6bc314360183cde0f155f, but without any test coverage or so.

Like in the other cases, this function never returns a `WP_Error`

mrsdizzie (approved):

PR #546, opened 2025-10-15
feat: add --all flag to comment delete command

### LINKED ISSUE #538: Add an `--all` flag to `wp comment delete`
## Feature Request

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

A website under my purview had hundreds of thousands of spam comments. It would be nice to be able to remove them in one fell swoop without having to resort to database queries (`truncate wp_comments; truncate wp_commentmeta;`).

I'd like to add a flag to the `wp comment delete` subcommand:

```
wp comment delete --all
```

This may have a confirmation prompt along the lines of "Are you sure you want to delete XXXXX comments? (Y/n)" There might also be a "--yes" option for non-interactive use:

```
wp comment delete --all --yes
```

The addition of the "--all" flag would mirror similar-ish flags for the `wp plugin delete --all` and `wp theme delete --all` commands.

I'm aware that the `wp site empty` command also removes all comments, but this is a little bit different.

I'd be happy to take a stab at adding this modification, if approved.

shawnhooper:
Resolves #538 

Contributed as part of WordCamp Canada 2025 Contributor Day.  #WCEH

### DIFF
--- a/src/Comment_Command.php
+++ b/src/Comment_Command.php
@@ -807,7 +807,7 @@ class Comment_Command extends CommandWithDBObject {
 	 * Gets all available comment IDs.
 	 */
 	private function get_all_comment_ids(): array {
-		$query    = new WP_Comment_Query();
+		$query = new WP_Comment_Query();
 
 		/** @var array $comments */
 		$comments = $query->query(


src/Comment_Command.php
```diff
@@ -734,4 +751,59 @@ public function exists( $args ) {
...
+
+	/**
+	 * Gets all available comment IDs.
+	 */
+	private function get_all_comment_ids(): array {
+		$query    = new WP_Comment_Query();
+		$comments = $query->query(
+			array(
+				'fields' => 'ids',
+				'number' => 0, // Get all comments
```
swissspidy, line 816: This sounds like it could be very slow on a large site

shawnhooper: Agreed, I didn't like the looping either, but I wanted it to still trigger the `delete_comment` and `deleted_comment` hooks. 

Do you have any suggestions?

swissspidy: Which makes sense I suppose, there is usually a lot attached to these filters. And `wp_delete_comment` itself does a lot as well, such as moving up children, deleting comment meta, and updating comment count. Plus, only this way can we adhere to the `--force` flag as well.

So yeah... there might be no way around it. Maybe it would be worth adding a warning to the docblock to indicate this might be slow if there are lots of comments.

We should probably use `wp_defer_comment_counting()` to defer updating comment counts, otherwise it will be done after every single deletion.

Something like:

```php
$defer_count = wp_defer_comment_counting();

wp_defer_comment_counting( true );

// delete all comments

wp_defer_comment_counting( $defer_count )
```

shawnhooper: @swissspidy Ok.  I've added in the support for `wp_defer_comment_counting`.   This got a little more complex, since the call to the `parent::_delete()` function exited WP-CLI after a single delete.

swissspidy: @shawnhooper That's amazing, thanks a lot for diving straight in!

I see you're battling with the Behat tests. Tip: If you are expecting an error output in a test, you need to use `When I try ...`

shawnhooper: @swissspidy Yep, that did it!  Thanks, was really banging my head against my desk.  Thinking it was STDOUT vs STDERR issues, etc.

src/Comment_Command.php
```diff
@@ -16,7 +16,7 @@
  *     # Update an existing comment.
  *     $ wp comment update 123 --comment_author='That Guy'
  *     Success: Updated comment 123.
- *
+
```
swissspidy: Looks like this needs to be undone.

src/Comment_Command.php
```diff
@@ -416,9 +416,12 @@ function ( $comment ) {
 	 *
 	 * ## OPTIONS
 	 *
-	 * <id>...
+	 * [<id>...]
 	 * : One or more IDs of comments to delete.
 	 *
+	 * [--all]
+	 * : If set, all comments will be deleted.
```
swissspidy: Should we warn that this might be slow?

src/Comment_Command.php
```diff
@@ -734,4 +777,48 @@ public function exists( $args ) {
...
+	private function get_all_comment_ids(): array {
+		$query    = new WP_Comment_Query();
+		$comments = $query->query(
+			array(
+				'fields' => 'ids',
+				'number' => 0, // Get all comments
+			)
+		);
+
+		if ( ! is_array( $comments ) ) {
```
swissspidy: I assume we need this only because of PHPStan? It should never really be an array. Alternatively we could add a `@var` annotation

shawnhooper: @swissspidy Yes, this was just for PHPStan.  I've replaced it with the @var annotation.

src/Comment_Command.php
shawnhooper: @swissspidy Just a warning in the comment like this?  Or do we need a warning on the prompt?

mrsdizzie:
> I propose two possible solutions:
> 
> 1. A separate `wp comment truncate` command
> 2. A `--skip-hooks` flag on the delete command, that if present with `--all` would execute the truncate commands for speed.

It needs to be separate from `delete`, because it will not `delete` anything in the existing sense. Perhaps `wp comment reset` or `wp comment empty` (similar to `wp site empty`) that would do the following:

```
truncate wp_comments; truncate wp_commentmeta;
wp comment recount $(wp post list --post_type=page,post --format=ids) // equivalent of this
```

With an appropriate description of what it does and when you'd want to use that.

PR #553, opened 2025-11-10
PHPStan fixes

swissspidy:
Addresses some newly reported PHPStan errors.

Helps unblock some of the other PRs.

### DIFF
--- a/src/Comment_Command.php
+++ b/src/Comment_Command.php
@@ -63,8 +63,11 @@ class Comment_Command extends CommandWithDBObject {
 	 *     # Create comment.
 	 *     $ wp comment create --comment_post_ID=15 --comment_content="hello blog" --comment_author="wp-cli"
 	 *     Success: Created comment 932.
+	 *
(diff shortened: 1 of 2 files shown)


src/Comment_Command.php
```diff
@@ -64,7 +64,7 @@ public function __construct() {
 	 *     $ wp comment create --comment_post_ID=15 --comment_content="hello blog" --comment_author="wp-cli"
 	 *     Success: Created comment 932.
 	 */
-	public function create( $args, $assoc_args ) {
+	public function create( $args, array $assoc_args ) {
```
schlessera: Typehinting against arrays is not a good idea in PHP, as it then breaks anything that is object-based and supposed to work in the same way, like an `ArrayAccess` implementation. I would avoid directly using the `array` typehint at all unless the entire point is to make sure it only works with the exact type "array" and not anything array-like.
We can still use docblocks and/or union types here.

swissspidy: Fixed using docblocks.

PR #624, opened 2026-07-22
Fix: warn on potentially misspelled parameters in wp user update

ikamal7:
## Description

When a typo'd flag like `--user-pass` is used instead of `--user_pass` in `wp user update`, the command now emits a warning with a suggestion instead of silently accepting the unknown parameter.

Fixes #5286

## Approach

Added a check in `User_Command::update()` that compares unknown parameters against known user fields using `Utils\\get_suggestion()`. If a close match (Levenshtein distance ≤ 2) is found, a warning is emitted and the unknown parameter is removed before passing to `wp_update_user()`.

Custom user meta fields that don't closely match known fields are still silently accepted for backward compatibility.

## Testing

- `wp user update 2 --user-pass=newpass` → `Warning: unknown --user-pass parameter (did you mean '--user_pass'?)`
- `wp user update 2 --user_pass=newpass` → `Success: Updated user 2.` (unchanged)
- `wp user update 2 --some_custom_field=value` → `Success` (no warning, BC preserved)

## Summary by CodeRabbit

* **New Features**
  * Added warnings for likely misspelled parameters when updating users.
  * Suggested corrections are provided for unrecognized user field options.

* **Bug Fixes**
  * Prevented unknown or misspelled parameters from being passed through during user updates.

### DIFF
--- a/src/User_Command.php
+++ b/src/User_Command.php
@@ -37,6 +37,47 @@ use WP_CLI\Utils;
  */
 class User_Command extends CommandWithDBObject {
 
+	/**
+	 * Known user fields accepted by wp_update_user().
+	 *
+	 * @var array
+	 */
+	const KNOWN_USER_FIELDS = [
+		'user_pass',
+		'user_nicename',
+		'user_url',
+		'user_email',
+		'display_name',
+		'nickname',
+		'first_name',
+		'last_name',
+		'description',
+		'rich_editing',
+		'syntax_highlighting',
+		'comment_shortcuts',
+		'admin_color',
+		'use_ssl',
+		'show_admin_bar_front',
+		'user_registered',
+		'locale',
+		'role',
+		'meta_input',
+	];
+
+	/**
+	 * Command flags that are not user fields.
+	 *
+	 * @var array
+	 */
+	const COMMAND_FLAGS = [
+		'skip-email',
+		'defer-term-counting',
+		'porcelain',
+		'format',
+		'fields',
+		'user_login',
+	];
+
 	protected $obj_type   = 'user';
 	protected $obj_fields = [
 		'ID',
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanjs for your PR!

Please see discussions at #528 and https://github.com/wp-cli/wp-cli/issues/5286 on possible approaches.

We need a solution that works for all entities, not just users, and can be updated witz new core repeases.

ikamal7:
Thanks for the feedback @swissspidy!

I read through the discussion on #528 and understand the concerns. A few thoughts/questions:

1. **Cross-entity approach** — I agree a uniform solution is ideal. Would the preferred direction be to:
   - **A)** Remove `--<field>=<value>` from each entity command's docblock and complete the field list (as mrsdizzie suggested)? This gives us free validation from the framework but loses support for custom meta fields.
   - **B)** Build a shared entity field registry (as you originally suggested on #5286) that could live in the framework and be referenced by all commands?
   - **C)** Something else?

2. **Custom fields** — For `wp_update_user()` / `wp_update_post()` / `wp_update_comment()`, plugins can add custom meta via filters like `wp_pre_insert_user_data` or `meta_input`. If we remove `--<field>=<value>`, those would break. Is that a concern, or is that considered out of scope?

3. **Next steps** — I'm happy to either adapt this PR to whichever approach the team prefers, or help scope out a broader solution. What would be most useful?

swissspidy:
Breaking existing functionality is usually not desireable.

I recommend continuing the discussion at https://github.com/wp-cli/wp-cli/issues/5286 before jumping straight into code. Maybe there are other options too.

swissspidy:
Closing this one for now in favor of discussion at https://github.com/wp-cli/wp-cli/issues/5286

